Issue #215, opened 2014-08-19
Notice: Undefined index: required in class-tgm-plugin-activation.php on line 652


Michi23:
Issue occurs in WP backend while WP_DEBUG is active

line 652 current:
if ( $plugin['required'] ) {

line 652 fix:
if ( isset( $plugin['required'] ) && $plugin['required'] ) {

:-)

PR #216, opened 2014-08-26
Notice no longer thrown when 'required' missing

maciej-gurban:
PHP Notice no longer thrown when plugin array's 'required' key is missing.


GaryJones:
Would this need to be `isset( $plugin['required'] ) && $plugin['required']` so if it was set to `false`, the condition would act logically?

GaryJones:
Actually, this was fixed in 9086b9ae37c19a357265a5c4ed78b1e5ae0951ec, reported in #215

PR #219, opened 2014-09-05
[Develop] Improve extensibility

jrfnl:
The changes in this PR allow for the class to be extended from a child class for a specific implementation.

The way the class was set up, extending from a child class would mean overloading a truck load of methods to get round the hard coded class name and some more convoluted coding to overrule the instance and the auto-instantiation of the class.
These changes take the pain out of extending the class and make it so this library can have a clean upgrade path even when you extend the class.

#### Details:
* Wrap instantiation in a hooked function (so it can be unhooked)
* Use the global to get to the instance.
This allows for an extended child class to set the global and for everything to still work provided the child class provides their own static `get_instance()` method which sets their own static `$instance` property.
* Force instantiation via `get_instance()` by making the constructor protected

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -167,9 +167,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
          *
          * @see TGM_Plugin_Activation::init()
          */
-        public function __construct() {
-
-            self::$instance = $this;
+        protected function __construct() {
 
             // Set the current WordPress version.
             global $wp_version;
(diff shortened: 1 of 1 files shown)


GaryJones:
I'm +1 for this.

class-tgm-plugin-activation.php
```diff
@@ -1034,8 +1032,8 @@ public function force_deactivation() {
          */
         public static function get_instance() {
 
-            if ( ! isset( self::$instance ) && ! ( self::$instance instanceof TGM_Plugin_Activation ) ) {
-                self::$instance = new TGM_Plugin_Activation();
+            if ( ! isset( self::$instance ) && ! ( self::$instance instanceof self ) ) {
+                self::$instance = new self();
```
GaryJones: Instead of having an untestable singleton class, how about we put a static into `load_tgm_plugin_activation()` instead?

GaryJones:
Applied manually in 7ce1ea0de1b35b714386926ac8e5505328731dcf. Thanks @jrfnl.

PR #221, opened 2014-09-07
[Develop] Turn parent menu slug into a config variable

jrfnl:
Allows for setting of the parent menu item from the $config array

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -86,6 +86,15 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
         public $menu = 'tgmpa-install-plugins';
 
         /**
+         * Parent menu file slug
+         *
+         * @since 2.4.x
+         *
+         * @var string
+         */
+        public $parent_slug = 'themes.php';
+
+        /**
          * Default absolute path to folder containing pre-packaged plugin zip files.
          *
          * @since 2.0.0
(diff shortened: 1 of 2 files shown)


jrfnl:
@thomasgriffin Any feedback on this ?

senlin:
This looks great, would be awesome if it can be integrated. Thanks!

jrfnl:
@senlin Glad you like it ;-)

@thomasgriffin I kind of have the feeling that you may be hesitant about this PR because it builds on the changes from @zackkatz's PR #188 which were merged in `develop`, but not in `master` because of the theme check disallowing the use of `add_submenu_page()`.

I think I can provide a solution.

A while ago, I send in PR #219 with some changes to allow for overloading of the class. If that PR would be merged, we can then leave the 'old' method in the base class which would be used by themes.
For plugins which may want to change where the menu is added, an overloaded class can be added which adjusts the `admin_menu()` method to use `add_submenu_page()` instead.

I might need to still make some small changes to the code in #219 to make this more intuitive for less experienced programmers, but would like your opinion before I spend a lot (more) time on something which may never get accepted.

So what do you think ?

jrfnl:
Also: the current getting of the admin page with the code in the `develop` branch what with the hard-coded `themes.php`, only works because of a recognized *bug* in WP. We can't count on the bug staying in.
Ref: https://core.trac.wordpress.org/ticket/28821

GaryJones:
+1 for abstracting `'themes.php'`.

0 (but willing to be swayed) for making them properties as well. What's the reasoning there? To be able to subclass? (If so, how does that handle with the main class being a singleton?)

jrfnl:
> 0 (but willing to be swayed) for making them properties as well. What's the reasoning there? To be able to subclass? (If so, how does that handle with the main class being a singleton?)

I followed the logic which was already in place - the `TGM_Plugin_Activation::config()` method currently sets all config variables passed as properties, i.e. the properties have to be there.

GaryJones:
+1 for being consistent then :-)

jrfnl:
Related #65

jrfnl:
Related 44228afebcae2128455801e154546b82706412fa and 5c76406bf515ec2a0919548fe63514b3b0b9131b which seem to be reversed/taken out by c44d6aa0fd9c8050f97bce5e5f96cc0f6b19c511

PR #227, opened 2014-09-24
Default WP_List_Table Styles Rule for UI Improvement

shivapoudel:
Since `TGMPA_List_Table` is hijacking and extending `WP_List_Table` which is private class and is Limited for core usage only. Similary, we are utilizing similar plural args of `WP_Plugins_List_Table`  too. Means we are adding plugins page style rules directly from the core plugins page and our table `<thead> and <tfoot>` checkbox are not well managed as well as other css stuffs. Thus let's be smart and accept default `WP_List_Table` styles rather than Plugins Page. 

``` php
/**
 * Confidential Information!
 * WP_Plugins_List_Table is already utilizing 'plural' => 'plugins'
 * Let's be smart dudes and use only singular args :)
 */
parent::__construct( array(
    'plural' => 'plugins',
    'screen' => isset( $args['screen'] ) ? $args['screen'] : null,
) );
```

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1115,7 +1115,6 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
             parent::__construct(
                 array(
                     'singular' => 'plugin',
-                    'plural'   => 'plugins',
                     'ajax'     => false,
                 )
             );


shivapoudel:
@thomasgriffin and @GaryJones can this be commited for master and othe 3.0.0 branch too ;)

jrfnl:
@shivapoudel It will definitely be included in the 3.0 branch, not to worry (not sure about master).

shivapoudel:
Cool! When is the release date?

jrfnl:
@shivapoudel eh... when it's ready... ;-)

GaryJones:
It won't be in 2.4.1 but it will be in 2.5.0

PR #246, opened 2015-01-13
[Develop] Bug fix: only the user which switched theme would be re-notified

jrfnl:
There may be several admin users for a site. If a theme using TGMPA is deactivated, the dismiss notice state would only be removed for the user doing the theme switching, not for all users.
This effectively meant that only that user would be re-notified of open recommended/required plugins if the theme would be reactivated.

This fix ensures that all eligible users will be re-notified on reactivation of the theme using TGMPA.


thomasgriffin:
Would you mind creating this pull request against the master branch? I've not done anything with the develop branch in a while. Indeed, I need to actually remove it as it doesn't quite make sense for this repo.

Issue #237, opened 2014-11-16
TGM notices appear twice on theme options pages


manake:
I'm using the newest v2.4.0.

```
function prefix_theme_options_add_page() {
	add_theme_page( __( 'Theme Options', 'textdomain' ), __( 'Theme Options', 'textdomain' ), 'edit_theme_options', 'theme_options', 'prefix_theme_options_render_page' );
}
add_action( 'admin_menu', 'prefix_theme_options_add_page' );

function prefix_theme_options_render_page() {
	?>
	<div class="wrap">
		<?php screen_icon(); ?>
		<h2><?php _e( 'Theme Options', 'textdomain' ); ?></h2>
		<?php settings_errors(); // This causes TGM class notices to appear twice. When I remove this I don't have "Page saved!" message generated by WordPress. ?>

		<form method="post" action="options.php">
			<?php
				settings_fields( 'my_options' );
				do_settings_sections( 'theme_options' );
				submit_button();
			?>
		</form>
	</div>
	<?php
}

// This notice never appears twice.
function my_admin_notice() {
    ?>
    <div class="updated">
        <p><?php _e( 'Updated!', 'textdomain' ); ?></p>
    </div>
    <?php
}
add_action( 'admin_notices', 'my_admin_notice' );
```

I think the `settings_errors( 'tgmpa' );` line should be replaced with something else in the plugin.

jrfnl:
I've send in PR #249 to fix this, please give it a +1 if you're happy with it.

PR #249, opened 2015-01-13
[Develop] Prevent duplicate message showing

jrfnl:
Quite a few themes and plugins which don't have their settings pages under 'options-general' also use `settings_errors()` to display their notices to the user.

This resulted in the TGMPA messages being shown twice on those pages as reported in #237.

This PR fixes that.

Related: https://core.trac.wordpress.org/ticket/31000


thomasgriffin:
Would you mind creating this pull request against the master branch? I've not done anything with the develop branch in a while. Indeed, I need to actually remove it as it doesn't quite make sense for this repo.

GaryJones:
Needs a `@since` tag, and should it be `protected` not `private`?

jrfnl:
@GaryJones Thanks for the feedback, adjusted. Not sure what the next version nr will be though, so left it at `2.4.x` for now.

PR #247, opened 2015-01-13
Fix: multisite admin menu and notices

jrfnl:
As on multisite by default plugins can only be installed from the network admin, it makes more sense showing the TGM page in the network admin and not in the site admin.

On that same note, the admin notices should be shown on the network admin too.

Related issues: #228, #239

Issue #239, opened 2014-12-03
Permission error on multisite setup when clicking 'Begin installing plugin' link


JohEngstrom:
I've found one issue.

In a multisite setup I've set an official repository plugin as required. The notice comes up in the admin but when clicking on the Begin installing plugin link I'm presented with "You do not have sufficient permissions to access this page.".

I am the only admin user on this WordPress install.

The URL the link goes to is /wp-admin/network/themes.php?page=tgmpa-install-plugins

jrfnl:
Probably duplicate of issue #228

Issue #228, opened 2014-09-26
WP4.0 MU i get permission error.


parkmap:
I got a "You do not have sufficient permissions to access this page." error on /wp-admin/network/themes.php?page=tgmpa-install-plugins

current version is wordpress  4.0 Multisite

jrfnl:
PR #247 applies this fix and also fixes the admin notice display for multisite.

CodingSamurai:
Thank you this saved me after a couple of hours of frustration!  This really should be rolled into V2.5 as a temporary fix until the restructured behaviour is built in 3.0

jrfnl:
@CodingSamurai While I understand your point, as MS is not supported at this moment, we shouldn't be giving the impression that it is, which is why this will not be included in v2.5.

quitoxic:
Hey,
I have no experience so far with PHP and getting this problem, same thing multisite etc. 
I have been trying to follow the instructions on this thread but faild to make it work. Could any one help me out?

jrfnl:
@quitoxic As indicated in the post just above yours: Multi-site is currently not supported by TGMPA nor could it be with the way it is set up at this moment. Making TGMPA multi-site compliant is one of our goals for version 3.0, but will involve major changes in how to indicate dependencies.

PR #252, opened 2015-01-13
[Master] Fix: multisite admin menu and notices

jrfnl:
As on multisite by default plugins can only be installed from the network admin, it makes more sense showing the TGM page in the network admin and not in the site admin.

On that same note, the admin notices should be shown on the network admin too.

Related #228, #239

Sister-PR to PR #247

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -314,8 +314,11 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
          */
         public function thickbox() {
 
-            if ( ! get_user_meta( get_current_user_id(), 'tgmpa_dismissed_notice', true ) ) {
-                add_thickbox();
+            if ( ! is_multisite() || ( is_multisite() && is_network_admin() ) ) {
+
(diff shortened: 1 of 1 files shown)


jrfnl:
On second though: This should not be merged as is. The logic is valid for TGM being included with a plugin, however, when TGM is enabled for a theme, the logic does not hold.

The problem then is that when on multisite:
- The theme used is site-specific, so whether TGM is active will differ per site.
- However plugins can normally only be installed from the network admin and not every site-admin has access to the network admin, which causes an issue.
- Similarly, as the theme loaded when on the network admin is the theme associated with site 1, it will be difficult to detect when to show a network admin a notice about installing plugins via TGM needed for themes in use by child-sites other than site 1.

Daronspence:
Is this problem going to be addressed? Or are MU users SOL?

jrfnl:
@Daronspence This will be addressed and probably quite soon too.

Daronspence:
@jrfnl I've used the `$ms_hook` variable you defined in one of your commits and it seems to be working nicely for my needs. I've got the class semi-heavily customized at this point so it's not a huge deal for me at this point, though I'm sure it would certainly benefit others!

Issue #171, opened 2014-03-17
Permission error on multisite


EusebiuOprinoiu:
Hello, Thomas!
It looks like there is a problem with the new <code>network_admin_url</code> instance on multisite. The plugins won't activate. This is the received error: "You do not have sufficient permissions to access this page." (even though the user is the network admin)
In version 2.3.6, where <code>admin_url</code> is used everything works the way it should..

PR #255, opened 2015-01-15
[Develop] Use `self_admin_url()` rather than `admin_url()`

jrfnl:
Not well documented, but has been around since WP 3.1
@see https://developer.wordpress.org/reference/functions/self_admin_url/

Refs: #171, 852c7e13e1255716f72f8e42ae02d306ab18b0dd


GaryJones:
+1.

PR #269, opened 2015-02-16
Escaping HTML attributes and localization fixes

webmandesign:
Escaping HTML title attribute and applying some localization fixes.


GaryJones:
Manually applied most of the fixes in https://github.com/thomasgriffin/TGM-Plugin-Activation/commit/633f1c107fc86e319693cc5a4111170b226910e7, thanks!

PR #281, opened 2015-03-28
Use EXTR_SKIP in extract function

Ninos:
You should use the ```EXTR_SKIP``` flag to avoid conflicts with already existing variables.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1812,7 +1812,7 @@ if ( ! class_exists( 'WP_Upgrader' ) && ( isset( $_GET['page'] ) && TGM_Plugin_A
 
                 // Parse default options with config options from $this->bulk_upgrade and extract them.
                 $options = wp_parse_args( $options, $defaults );
-                extract( $options );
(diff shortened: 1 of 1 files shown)


jrfnl:
As the extract is used within the scope of a function, except for the `$options` and `$defaults` variables, no other variables exist and therefore there is no change of conflict...

Ninos:
I would use it per default. May in near future you use a global variable and forget to add this flag. With this pr you avoid possible future conflicts..

GaryJones:
Thanks @Ninos.

The approach within WP is to avoid `extract()` altogether. Here, just using `$options['somekey']` can be used instead of `$somekey`.

As such, I'm going to close this PR, but I'm open for another one that removes all instances of `extract()`.

Ninos:
PS: @GaryJones can you may send me a link to the decision? Because I also use the extract function for all of my custom shortcodes (thought it is the common way). So may I'll change it too, if the wp-devs don't want to see this function anymore :D

GaryJones:
https://core.trac.wordpress.org/ticket/22400

Ninos:
Thx :)

PR #291, opened 2015-04-21
Bump version to 2.4.1 and update changelog

Stephen-Cronin:
Have bumped version to 2.4.1 and updated the change log to include both today's change of escaping add_query_arg and the previous change of removing call to deprecated `function screen_icon()`


Stephen-Cronin:
Okay - this was based on the version with the error, so I'm closing this....

PR #304, opened 2015-04-22
[Develop] Add travis script

jrfnl:
Needs IRL test

Also:
* Rename `phpcs.xml` to `ruleset.xml` so PHPStorm can pick it up
* Fix all but one remaining code style issue
* only show `var_dump()` on error when `WP_DEBUG` is set to `true`
* Remove unnecessary concatenation in echo statements for lower memory usage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,47 @@
+# Travis CI (MIT License) configuration file
+# @link https://travis-ci.org/
+
+# Use new container based environment
+sudo: false
+
+# Declare project language.
+# @link http://about.travis-ci.org/docs/user/languages/php/
+language: php
+
+# Declare versions of PHP to use. Use one decimal max.
+php:
+    # aliased to a recent 5.5.x version
+    - "5.5"
+    # Lowest currently possible in travis
+    - "5.2"
+
+before_script:
+    # Install CodeSniffer for WordPress Coding Standards checks.
+    - git clone https://github.com/squizlabs/PHP_CodeSniffer.git phpcs
+    # Install WordPress Coding Standards.
+    - git clone -b master https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git wpcs
+    # Hop into CodeSniffer directory.
+    - cd phpcs
+    # Set install path for WordPress Coding Standards.
+    # @link https://github.com/squizlabs/PHP_CodeSniffer/blob/4237c2fc98cc838730b76ee9cee316f99286a2a7/CodeSniffer.php#L1941
+    - scripts/phpcs --config-set installed_paths ../wpcs
+    # Hop into project dir.
+    - cd $TRAVIS_BUILD_DIR
+    # After CodeSniffer install you should refresh your path.
+    - phpenv rehash
+
+# Run test script commands.
+# All commands must exit with code 0 on success. Anything else is considered failure.
+script:
+    # Search for PHP syntax errors. Ignore the phpcs and wpcs directories.
+    - find -L . -path "./*pcs" -prune -o -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l;
+    # WordPress Coding Standards
+    # @link https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
+    # @link http://pear.php.net/package/PHP_CodeSniffer/
+    # -p flag: Show progress of the run.
+    # -s flag: Show sniff codes in all reports.
+    # -v flag: Print verbose output.
+    # -n flag: Do not print warnings (shortcut for --warning-severity=0)
+    # --standard: Use WordPress as the standard.
+    # --extensions: Only sniff PHP files.
+    - $TRAVIS_BUILD_DIR/phpcs/scripts/phpcs -p -s -v -n . --standard=./ruleset.xml --extensions=php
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -0,0 +1,49 @@
...
+language: php
+
+# Declare versions of PHP to use. Use one decimal max.
+php:
+    # Aliased to a recent high PHP version
+    - 5.6
+    # Might as well test HHVM too
+    - hhvm
+    # Lowest currently possible in travis
+    - 5.2
```
GaryJones: Doesn't Travis only support 5.3 min now?

What about 5.3, 5.4, 5.5? Occasionally something breaks there, and currently we would only know that it passed the low PHP level, but failed at 5.6.

class-tgm-plugin-activation.php
```diff
@@ -517,7 +517,12 @@ protected function do_plugin_install() {
 					$api = plugins_api( 'plugin_information', array( 'slug' => $plugin['slug'], 'fields' => array( 'sections' => false ) ) );
 
 					if ( is_wp_error( $api ) ) {
-						wp_die( $this->strings['oops'] . var_dump( $api ) );
+						if ( WP_DEBUG === true ) {
```
GaryJones: Is `WP_DEBUG` defined by WP as false if not defined in wp-config.php, or do we need a `defined()` check in here first?

class-tgm-plugin-activation.php
```diff
@@ -558,12 +563,12 @@ protected function do_plugin_install() {
 					$this->populate_file_path(); // Re-populate the file path now that the plugin has been installed and activated.
 
 					if ( is_wp_error( $activate ) ) {
-						echo '<div id="message" class="error"><p>' . $activate->get_error_message() . '</p></div>';
-						echo '<p><a href="' . esc_url( add_query_arg( 'page', urlencode( $this->menu ), self_admin_url( 'themes.php' ) ) ) . '" target="_parent">' . $this->strings['return'] . '</a></p>';
+						echo '<div id="message" class="error"><p>', wp_kses_post( $activate->get_error_message() ), '</p></div>',
+							'<p><a href="', esc_url( add_query_arg( 'page', urlencode( $this->menu ), self_admin_url( 'themes.php' ) ) ), '" target="_parent">', esc_html( $this->strings['return'] ), '</a></p>';
```
GaryJones: Excellent - switching echo calls to use commas instead of concat was going to be one of my next tasks :-)

ruleset.xml
```diff
@@ -5,5 +5,7 @@
 	<file>example.php</file>
 	<arg name="report" value="full"/>
 	<arg value="sp"/>
-	<rule ref="WordPress"/>
+	<rule ref="WordPress-Extra">
+		<exclude name="Generic.Files.LineEndings.InvalidEOLChar" />
```
GaryJones: Why this exclusion?

ruleset.xml
```diff
@@ -5,5 +5,7 @@
 	<file>example.php</file>
 	<arg name="report" value="full"/>
 	<arg value="sp"/>
-	<rule ref="WordPress"/>
```
GaryJones: Why removal of the overall WordPress ruleset?

GaryJones:
I think the addition of `.travis.yml`, the change in phpcs config file, and the code fixes should all have been in different branches. I like the code changes, but have questions about the other two.

.travis.yml
jrfnl: I'm pretty sure they reverted the removal of PHP 5.2 (under pressure of the WP community).

As we're currently only doing a lint and phpcs, running it on more versions would be wasting resources as the results won't change. If it breaks on either of those two versions, we need to look at the code, everything in between is caught by this.

Once we have unit tests, we should start running it on more versions, though still only lint and phpcs on highest and lowest.

class-tgm-plugin-activation.php
jrfnl: No need, WP defines it as `false` if not set: https://core.trac.wordpress.org/browser/trunk/src/wp-includes/default-constants.php#L59

ruleset.xml
jrfnl: Actually true, I'll revert that. It will break on my local machine, but I can use an adjusted ruleset locally.

ruleset.xml
jrfnl: Because I'm presuming that TGMPA will not be allowed to be used in VIP context and therefore (most of) the VIP rules don't apply.

jrfnl:
> I think the addition of .travis.yml, the change in phpcs config file, and the code fixes should all have been in different branches. I like the code changes, but have questions about the other two.

I'll keep that in mind for that future.

In this case I think they are related (except for the concatenation commit which was just extra). Adding the travis script which runs phpcs and will fail straight away on errors would mean that the travis config could not be merged. (providing we'll hook it in now to test this PR before merge)

ruleset.xml
jrfnl: Oh and that's quite apart still from the buggyness of some of the VIP rules, such as https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/186 which will break our build for sure.

jrfnl:
Updated PR

jrfnl:
@GaryJones shouldn't we hook travis in first to test this PR ?

GaryJones:
It will get tested once we are hooked up - we can adjust it from there.

PR #306, opened 2015-04-22
Argh... make it work with PHP 5.2 which can't deal with variable class names

jrfnl:
(Also: fixed two pseudo @since tags)


jrfnl:
Reverted as for some reason the admin page is now no longer accessible. Will fix that first and send in new PR once fixed.

See https://github.com/thomasgriffin/TGM-Plugin-Activation/issues/303#issuecomment-95239997

PR #307, opened 2015-04-22
Introduce filter tgmpa_configuration settings

shivapoudel:
Let's introduce filter for this configuration xD


GaryJones:
The strings are part of the config, which you pass to `tgmpa()` when setting up the registration of plugins. No filter is needed.

PR #309, opened 2015-04-22
Really fix php5.2 compat

jrfnl:
Ok, found the typo - not sure about my head.
Tested & found working.


GaryJones:
I much prefer usage of `$this->tgmpa` where it's been set via the constructor, rather than `$GLOBALS` everywhere. It's one stage closer to being an injected dependency.

Excellent.

jrfnl:
@GaryJones I concur - pretty sure I actually coded that out at some point, not sure anymore why I changed it. Anyway, done properly now.

PR #321, opened 2015-04-23
Fix issues with bulk installing pre-packaged/bundled plugins.

### LINKED ISSUE #316: Bulk install issue with bundled plugins - develop branch
Hello,

Sorry if this has been previously notified and I couldn't find.

<strong>Case 1:</strong>  Bulk install of bundled plugins only

case-1

<strong>Case 2:</strong>  Bulk install of - Bundled Plugins + Plugins via WP repository 

case-2

Thanks, Tansh

jrfnl:
Fixed #316

Regression introduced in v2.4.1


GaryJones:
Can we have some inline comments as to why we escape one but encode the other please?

jrfnl:
Done

PR #322, opened 2015-04-23
Update file permissions

GaryJones:


jrfnl:
What's the real issue here ? Doesn't WP set this itself when installing themes ?

Also: Merge conflict

shivapoudel:
@GaryJones How can I set file permission using git bash? If possible help me please :sparkling_heart:

GaryJones:
> Doesn't WP set this itself when installing themes

Not that I know of - if someone copies and pastes the theme folder in, there's no file handling by WP whatsoever. We know it should be 644, so let's make it that instead of 755 which these two files happened to be (example.php was already 644).

@shivapoudel Standard command line: `chmod 644 my-file-name.php`. Once staged (`git add`), you can use `git status` to see the old and new permissions.

GaryJones:
Unconflicted.

shivapoudel:
@GaryJones Thanks!

PR #323, opened 2015-04-23
Add back some escapes which were added in 2.4.1

jrfnl:
but not carried over to develop or possibly accidentally removed in subsequent merges.


GaryJones:
Most likely I missed them.

PR #324, opened 2015-04-23
Escape some more output

jrfnl:
And found some more output which needs escaping


GaryJones:
Conflicting.

jrfnl:
Conflicts solved (was due to the other escaping PR)

PR #325, opened 2015-04-23
Only flush the plugins cache, not *all* WP caches.

jrfnl:
This will save a lot of overhead of other (page/metadata) caches having to be recreated because we flushed them.

This function was introduced in WP 3.7, so requires upping the minimum requirement of TGMPA.

Alternatively - if we wouldn't want to up the minimum requirement yet -, we could wrap this in a `function_exists()` if then else block.

[Edit]
Looking into this further, we could also choose to remove the flush calls completely and just up the minimum requirements as a quick check shows that WP 3.7+ will flush the plugin cache already. This, however, would need to be thoroughly checked for all situations in which we flush the cache.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 **Contributors:** Thomas Griffin (@jthomasgriffin / thomasgriffinmedia.com), Gary Jones (Github: @GaryJones / Twitter: GaryJ)  
 **Version:** 2.4.1  
-**Requires at least:** 3.0.0  
+**Requires at least:** 3.7.0  
 **Tested up to:** 4.2  
 
 ## Description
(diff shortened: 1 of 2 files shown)


GaryJones:
I'm fine with upping the minimum level. I don't think anyone (TGMPA or themes and plugins that use it) should be supporting that far back anyway. Now 4.2 is out, I'd rather say that 4.0 is the minimum.

Leaving open until you're happy enough to add the reviewmerge label and milestone.

jrfnl:
I'm happy to merge it like this for the 2.5 release. I suggest we revisit if we still need the flushing at all when developing 3.0.

PR #327, opened 2015-04-24
Adjust plugin directory name for "wrongly" packaged zips.

### LINKED ISSUE #280: Failing to rename temporary folders
Im using https://github.com/lmartins/webipack-posts-widget/raw/master/webipack-posts-widget.zip as the source parameter for one plugin I need. The plugin is downloaded and I get the message everything is ok. However when I go to the plugins page I get the notice the plugin was disabled for missing files.
Looking at the plugins folder, I see that the installed plugin folder has the name `webipack-posts-widget.tmp`. Im guessing something is failing and the `.tpm`never gets removed from that folder name.

Anything I can do about it?

### LINKED ISSUE #283: Name of plugin from GitHub
Hi,

I'm installing plugin from GitHub with TGM. After installation it adds '_master' word to plugin name, e.g. 'plugin-name_master'.
Is there any option to remove this extra word 'master'?

Thanks!

jrfnl:
This allows for instance for GitHub pre-packaged zips to be correctly installed and recognized.

Previously if you'd use the GitHub packaged zips - for instance: either master: https://github.com/lmartins/webipack-posts-widget/archive/master.zip or release based: https://github.com/lmartins/webipack-posts-widget/archive/v1.1.zip, you'd end up with an installed plugin, but with the plugin directory having a non-standard name `webipack-posts-widget-master`, cq `webipack-posts-widget-1.1` which would cause TGMPA not to recognize it as installed.

Fixes #280, Fixes #283

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -652,6 +652,14 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				return $source;
 			}
 
+			// Check for single file plugins
+			$source_files = array_keys( $GLOBALS['wp_filesystem']->dirlist( $remote_source ) );
+			if ( 1 === count( $source_files ) && false === $wp_filesystem->is_dir( $source ) ) {
+
+				return $source;
+			}
+
+			// Multi-file plugin, let's see if the directory is correctly named
 			$desired_slug = '';
 
 			// Figure out what the slug is supposed to be
@@ -669,19 +677,21 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			}
 
 			if ( '' !== $desired_slug ) {
-
 				$subdir_name = substr( str_replace( $remote_source . '/', '', $source ), 0, -1 );
 
-				if ( $subdir_name !== $desired_slug ) {
+				if ( ! empty( $subdir_name ) && $subdir_name !== $desired_slug ) {
 					$from = substr( $source, 0, -1 ); // remove end slash
-					$to   = $remote_source . '/' . $desired_slug;
+					$to   = trailingslashit( $remote_source ) . $desired_slug;
 
 					if ( true === $GLOBALS['wp_filesystem']->move( $from, $to ) ) {
-						return $to . '/';
+						return trailingslashit( $to );
 					}
 					else {
 						return new WP_Error( 'rename_failed', __( 'The remote plugin package is does not contain a folder with the desired slug and renaming did not work. Please contact the plugin provider and ask them to package their plugin according to the WordPress guidelines.', 'tgmpa' ), array( 'found' => $subdir_name, 'expected' => $desired_slug ) );
 					}
+				}
+				elseif ( empty( $subdir_name ) ) {
+						return new WP_Error( 'packaged_wrong', __( 'The remote plugin package consists of more than one file, but the files are not packaged in a folder. Please contact the plugin provider and ask them to package their plugin according to the WordPress guidelines.', 'tgmpa' ), array( 'found' => $subdir_name, 'expected' => $desired_slug ) );
 				}
 			}
 
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -552,7 +552,9 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				$upgrader = new Plugin_Upgrader( $skin = new Plugin_Installer_Skin( compact( 'type', 'title', 'url', 'nonce', 'plugin', 'api' ) ) );
 
 				// Perform the action and install the plugin from the $source urldecode().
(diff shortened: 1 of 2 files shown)


class-tgm-plugin-activation.php
```diff
@@ -630,6 +632,73 @@ protected function do_plugin_install() {
...
+		 */
+		public function maybe_adjust_source_dir( $source, $remote_source, $upgrader ) {
+
+			if ( ! $this->is_tgmpa_page() ) {
+				return $source;
+			}
+
+			// Check for single file plugins
+			$source_files = array_keys( $GLOBALS['wp_filesystem']->dirlist( $remote_source ) );
+			if ( 1 === count( $source_files ) && false === $wp_filesystem->is_dir( $source ) ) {
```
GaryJones: Does there need to be a `global $wp_filesystem`, or is it already effectively done so by the `$GLOBALS['wp_filesystem`] on the previous line?

class-tgm-plugin-activation.php
```diff
@@ -630,6 +632,73 @@ protected function do_plugin_install() {
...
+				foreach ( $this->plugins as $plugin ) {
+					if ( $plugin['name'] === $upgrader->skin->plugin_names[ $upgrader->skin->i ] ) {
+						$desired_slug = $plugin['slug'];
+						break;
+					}
+				}
+			}
+
+			if ( '' !== $desired_slug ) {
+				$subdir_name = substr( str_replace( $remote_source . '/', '', $source ), 0, -1 );
```
GaryJones: Better to use a `trailingslashit()` here instead of concat `/`?

class-tgm-plugin-activation.php
```diff
@@ -630,6 +632,73 @@ protected function do_plugin_install() {
...
+
+				if ( ! empty( $subdir_name ) && $subdir_name !== $desired_slug ) {
+					$from = substr( $source, 0, -1 ); // remove end slash
```
GaryJones: `untrailingslashit()` would be better here.

class-tgm-plugin-activation.php
```diff
@@ -630,6 +632,73 @@ protected function do_plugin_install() {
...
+					$to   = trailingslashit( $remote_source ) . $desired_slug;
+
+					if ( true === $GLOBALS['wp_filesystem']->move( $from, $to ) ) {
+						return trailingslashit( $to );
+					}
+					else {
+						return new WP_Error( 'rename_failed', __( 'The remote plugin package is does not contain a folder with the desired slug and renaming did not work. Please contact the plugin provider and ask them to package their plugin according to the WordPress guidelines.', 'tgmpa' ), array( 'found' => $subdir_name, 'expected' => $desired_slug ) );
```
GaryJones: Should this string (and the one below) be moved with the rest of the strings and then simply referenced here?

class-tgm-plugin-activation.php
jrfnl: Yes and no. Good catch - I copied that line from the WP_Upgrader class and obviously missed that. Fixed.

class-tgm-plugin-activation.php
jrfnl: Agreed

class-tgm-plugin-activation.php
jrfnl: Agreed

class-tgm-plugin-activation.php
jrfnl: Actually there are several places were we have 'loose' strings. Also there are three places were we define strings in bulk - the `TGM_Plugin_Activation` class,  the `TGM_Bulk_Installer` class and the `TGM_Bulk_Installer_Skin` class. The _most_ logical place for these strings would be in `TGM_Bulk_Installer`, but the new function is not part of the bulk installer class which makes it a bit awkward.
The reason it's not part of the class is that with the way things are set up now, we don't have a separate TGM_Plugin_Upgrader (between Plugin_Upgrader and Bulk) - we use the normal Plugin_upgrader for single plugin installs (from the TGMPA class) and a custom bulk upgrader for bulk. So the function has to be in the TGMPA class, but the strings do not fit there properly....

I suggest we leave this for now and revisit when all the classes get refactored anyway in v3.

jrfnl:
Updated

GaryJones:
> Because the second part is the same in both messages, so makes for easier and more consistent translations.

Logical :+1:

Once we split the files up, I'd like to look into having something along the lines of a `TGMPA_Strings` class which holds all of the strings, and could be accessed either statically or via dependency injection etc.

Merged.

jrfnl:
> Once we split the files up, I'd like to look into having something along the lines of a TGMPA_Strings class which holds all of the strings, and could be accessed either statically or via dependency injection etc.

We'll need to think about this carefully. Once we make the strings more context-agnostic, there should be little need for clients to overrule our strings.
Also it would be nightmarish to deal with plugin A overruling the strings, then theme B overruling the strings and us getting bug reports about error messages we don't recognize.

PR #330, opened 2015-04-24
Add Scrutinizer config file

GaryJones:
https://scrutinizer-ci.com

Tested against a fork of this repo (resulting suggestions to be applied later on) so will work when this repo is hooked up too.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,0 +1,50 @@
+checks:
+    php:
+        use_self_instead_of_fqcn: true
+        uppercase_constants: true
+        simplify_boolean_return: true
+        return_doc_comment_if_not_inferrable: true
+        remove_extra_empty_lines: true
+        prefer_while_loop_over_for_loop: true
+        parameter_doc_comments: true
+        param_doc_comment_if_not_inferrable: true
+        optional_parameters_at_the_end: true
+        no_short_method_names:
+            minimum: '3'
+        no_long_variable_names:
+            maximum: '25'
+        no_goto: true
+        newline_at_end_of_file: true
+        naming_conventions:
+            local_variable: '^[a-z][a-z_0-9]*$'
+            abstract_class_name: ^Abstract|Factory$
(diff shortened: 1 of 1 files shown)


jrfnl:
Is there anything we can do about the 'else on same line as brace from previous if' ? It's not part of the official WPCS and having it on the next line make for easier to read code.

GaryJones:
> Is there anything we can do about the 'else on same line as brace from previous if' ? It's not part of the official WPCS and having it on the next line make for easier to read code.

It should be `} else {` as it is [part of the standard](https://make.wordpress.org/core/handbook/coding-standards/php/#brace-style). Scrutinizer does have a check for it, so we should be matching WPCS, i.e. your
```
}
else {
```
are not correct.

jrfnl:
> It should be } else { as it is part of the standard. 

Just read through it again and no, it does not say that. All it says is that braces should be used (rightfully so), not whether the else *should* be on the same or the next line.

shivapoudel:
@jrfnl IMO having `} else {` look pretty good rather than your's 
```
}
else {
```

GaryJones:
It does say that: "Braces shall be used for all blocks *in the style* shown here". All the examples on that page use it, and most, if not all instances in WP core use it. It may not explicitly say "Put the `}` on the same line as the `else {`, but it's a fairly standard to use the One True Brace Style (variant of K&R) in WP. PSR-2 is the same for control structures.

jrfnl:
@GaryJones Darn, did I really read over that bit ? Ok, never mind, the standard it is. Just weirdly inconsistent with the rest of the standard (less whitespace while most of the standard goes for more whitespace).

@shivapoudel I didn't say anything about looking 'pretty/good'. My point was about ease of readability.

jrfnl:
@GaryJones oh and are you happy with the resulting scrutinizer behaviour ? I.e. can I merge ?

GaryJones:
Merge as is please - we can pick up issues about code style or other config tweaks later on.

PR #337, opened 2015-04-24
EditorConfig + Contribution Guidelines

shivapoudel:
:gift:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,9 @@
+# This file is for unifying the coding style for different editors and IDEs
 # editorconfig.org
+
+# WordPress Coding Standards
+# http://make.wordpress.org/core/handbook/coding-standards/
+
 root = true
 
 [*]
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,29 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 4
+tab_width = 4
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,13 +1,13 @@
...
 
-# Use new container based environment
-sudo: false
-
 # Declare project language.
 # @link http://about.travis-ci.org/docs/user/languages/php/
 language: php
 
+# Use new container based environment
+sudo: false
```
GaryJones: Why do it later? Logically we would want to know where we're going to run stuff, before knowing what we're going to run?

.editorconfig
```diff
@@ -0,0 +1,34 @@
...
+indent_style = tab
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+indent_style = space
+indent_size = 2
+
+[*.txt]
```
GaryJones: We don't have .txt or .json files, so these can be removed.

.editorconfig
```diff
@@ -0,0 +1,34 @@
...
+trim_trailing_whitespace = false
+
+[*.json]
+insert_final_newline = false
+indent_style = space
+indent_size = 2
+
+[*.yml]
```
GaryJones: Since this is the same as Markdown, you can change line 18 to `[{*.md,*.yml}]` and delete this block.

GaryJones:
Please do these as separate branches / PRs, so if we need to revert one, we don't have to revert unrelated bits.

One for editorconfig, one for readme, etc.

And please squash the commits before pushing.

shivapoudel:
Close in favor of #339 and #340

PR #339, opened 2015-04-24
Configure .editorconfig for IDE

shivapoudel:
@GaryJones this not work very well:

```
[{*.md,*.yml}]
```

Usually works with full names like:

```
[{README.md,.coveralls.yml,.etc.yml}]
```

I'm using with VIM and Sublime and does not work. Impossible to use something unstable only for satisfaction of saving three lines in a configuration file.
;)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,9 @@
+# This file is for unifying the coding style for different editors and IDEs
 # editorconfig.org
+
+# WordPress Coding Standards
+# http://make.wordpress.org/core/handbook/coding-standards/
+
 root = true
 
 [*]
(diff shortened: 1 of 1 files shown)


GaryJones:
I'm only going by what I see on http://develop.svn.wordpress.org/trunk/.editorconfig .

Try `[*.{md,yml}]` instead, as that's the example given on http://editorconfig.org/

GaryJones:
Thanks @shivapoudel!

shivapoudel:
Markdown `trim_trailing_whitespace` should be false xD

For instance: We have Lead Developer and there at the end we have added the trailing whitespace for newline trick ;)

GaryJones:
I know. My ST3 is already setup with custom rules for .md files :-)

When I commented, it was against a line that said no newline at the end of the file, not trimming whitespace at the end of the line.

shivapoudel:
xD

PR #346, opened 2015-04-25
Improve code quality

GaryJones:
Fix some PHPCS and Scrutinizer issues.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -249,7 +249,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			// After this point, the plugins should be registered and the configuration set.
 
 			// Proceed only if we have plugins to handle.
-			if ( $this->plugins ) {
+			if ( ! empty( $this->plugins ) ) {
 				$sorted = array();
 
 				foreach ( $this->plugins as $plugin ) {
@@ -321,7 +321,8 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				wp_enqueue_style( 'plugin-install' );
 
 				global $tab, $body_id;
-				$body_id = $tab = 'plugin-information';
+				$body_id = 'plugin-information';
+				$tab     = 'plugin-information';
 
 				install_plugin_information();
 
@@ -480,9 +481,9 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			if ( isset( $_GET['plugin'] ) && ( isset( $_GET['tgmpa-install'] ) && 'install-plugin' === $_GET['tgmpa-install'] ) ) {
 				check_admin_referer( 'tgmpa-install' );
 
-				$plugin['name']   = $_GET['plugin_name']; // Plugin name. // @todo needs sanitizing, figure out how
-				$plugin['slug']   = sanitize_title( $_GET['plugin'] ); // Plugin slug.
-				$plugin['source'] = $_GET['plugin_source']; // Plugin source. // @todo needs sanitizing, figure out how
+				$plugin['name']   = $_GET['plugin_name'];
+				$plugin['slug']   = sanitize_title( $_GET['plugin'] );
+				$plugin['source'] = $_GET['plugin_source'];
 
 				// Pass all necessary information via URL if WP_Filesystem is needed.
 				$url = wp_nonce_url(
@@ -498,6 +499,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 					),
 					'tgmpa-install'
 				);
+
 				$method = ''; // Leave blank so WP_Filesystem can populate it as necessary.
 				$fields = array( 'tgmpa-install' ); // Extra fields to pass to WP_Filesystem.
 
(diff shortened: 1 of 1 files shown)


class-tgm-plugin-activation.php
```diff
@@ -249,7 +249,7 @@ public function init() {
 			// After this point, the plugins should be registered and the configuration set.
 
 			// Proceed only if we have plugins to handle.
-			if ( $this->plugins ) {
+			if ( ! empty( $this->plugins ) ) {
 				$sorted = array();
```
jrfnl: Better:
```php
if ( is_array( $this->plugins ) && array() !== $this->plugins )
```

class-tgm-plugin-activation.php
```diff
@@ -862,12 +864,12 @@ public function notices() {
...
 						'install'  => ( current_user_can( 'install_plugins' ) )  ? $show_install_link  : '',
 						'activate' => ( current_user_can( 'activate_plugins' ) ) ? $show_activate_link : '',
-						'dismiss'  => $this->dismissable ? '<a class="dismiss-notice" href="' . esc_url( add_query_arg( 'tgmpa-dismiss', 'dismiss_admin_notices' ) ) . '" target="_parent">' . esc_html( $this->strings['dismiss'] ) . '</a>' : '',
+						'dismiss'  => $this->dismissable ? '<a href="' . esc_url( add_query_arg( 'tgmpa-dismiss', 'dismiss_admin_notices' ) ) . '" class="dismiss-notice" target="_parent">' . esc_html( $this->strings['dismiss'] ) . '</a>' : '',
 					)
 				);
 
 				$action_links = array_filter( $action_links ); // Remove any empty array items.
-				if ( $action_links ) {
+				if ( ! empty( $action_links ) ) {
```
jrfnl: dito

class-tgm-plugin-activation.php
```diff
@@ -1220,7 +1218,7 @@ function tgmpa( $plugins, $config = array() ) {
 			call_user_func( array( $instance, 'register' ), $plugin );
 		}
 
-		if ( $config ) {
+		if ( ! empty( $config ) ) {
```
jrfnl: dito

class-tgm-plugin-activation.php
GaryJones: Why is `array() !== $this->plugins` better than `! empty( $this->plugins )`?

jrfnl: Where do I start ? Empty is hardly *ever* the correct function to use.

1. empty does not test for type while you need an array two lines later for the `foreach` which will throw a notice if it doesn't get one
2. empty will pass on a non-empty string, boolean true, a non-zero integer and float etc
3. empty actually doesn't pass on the string '0' (not relevant in this case, but confuses people often)
etc...

Always test specifically for what you want to know - here: do we have a non-empty array. And in this case neither `if ( $this->plugins )` nor `if ( ! empty( $this->plugins ) )` does.

See: http://phpcheatsheets.com/test/

GaryJones: 1. and 2. - I'm not asking why the need for `is_array()`, I'm asking why you're checking against exactly `array()` instead of using `empty()`. By the time we get to that check, we've already established we're looking at it being an array, so 2. is irrelevant.

[Performance wise](http://maettig.com/code/php/php-performance-benchmarks.php#empty), `empty()` seems to be quicker than `=== array()`.

`empty()` is also what Scrutinizer recommended so I would expect they've done their homework.

`empty()` is also slightly more self-documenting than `=== array()`.

I'm happy to add the `is_array()` check, but the `empty()` should be used over `=== array()` IMO i.e. `if ( is_array( $this->plugins ) && ! empty( $this->plugins ) )`

jrfnl: As long as you first check for `is_array()`, it doesn't matter whether you use `$var !== array()`, `! empty( $var )` or `count( $var ) > 0` for the result.

Using `count()` is definitely slowest, so either of the first two is better. I seemed to remember that `!==array()` was slightly faster (might depend on the PHP version), but am happy to accept `empty()` just as well.

GaryJones: There's no technical reason for `is_array()` being first, and according to the bottom of http://www.phpbench.com/, `is_array()` is relatively expensive to check compared to seeing if it's some type that is empty. 

So: "Does this variable have a non-nil value, and is it an array?" may be quicker to abort for bad values compared to "Is this variable an array, and does it have a non-nil value?"

Of course, this is a micro-optimisation, so I'm happy with the already committed order.

jrfnl: As you say, micro-optimization as both will normally need to pass anyway.

class-tgm-plugin-activation.php
jrfnl: Mind changing this one as well ? Will merge after that.

GaryJones: Thought I had, but it was another one I'd changed.

GaryJones:
Changed line 873/4 to be a cast to an array, since array_filter() was being called on it before checking it was not empty.

class-tgm-plugin-activation.php
```diff
@@ -1757,7 +1760,7 @@ public function process_bulk_actions() {
 				unset( $plugin, $temp );
 
 				if ( is_wp_error( $api ) ) {
-					if ( WP_DEBUG === true ) {
+					if ( true === WP_DEBUG ) {
```
shivapoudel: Why not `defined( 'WP_DEBUG' )` ?

GaryJones: Because I could have `define( 'WP_DEBUG', false )`.

WP already defines it as false if we've not defined it as something else, so the `WP_DEBUG` constant will always be defined - we just want to check it's been defined as true.

PR #348, opened 2015-04-25
Add update mode

swiderski:
We have merged current version of TGM plugin with version that provides updates for plugins.

PR #349, opened 2015-04-26
Added git ignorance file

shivapoudel:
Configured for Different Editors, OS X metadata and Windows Junk xD

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,15 @@
+# Editors
+project.xml
+project.properties
+/nbproject/private/
+.buildpath
+.project
+.settings*
+sftp-config.json
+.idea
+
+# OS X metadata
+.DS_Store
+
+# Windows junk
+Thumbs.db


GaryJones:
Folks really should have these setup as global gitignores on their own system, so I'm *meh* about having this. I'd rather it was just used to ignore project-related files, which we may not have any yet.

shivapoudel:
We can understand but this is for those for those folks we are unknown about global ignorance of project related files. For eg: #348 has submit the `.idea/misc.xml` file accidently :sweat_drops:

GaryJones:
I guessed where the inspiration comes from, but really, they should fix it once in their own system, than all of the repos containing code to ignore files that some people will never have (i.e. as a Mac user, I'm never going to be submitting Windows' Thumbs.db in PRs; I don't use Netbeans, so that line is redundant too).

jrfnl:
@shivapoudel FYI - You may find this article interesting to read:
http://365git.tumblr.com/post/519016351/three-ways-of-excluding-files

I have global .gitignore file with persistent thing set up and per project I may ignore some specific files which I add to /.git/info/excludes

PR #361, opened 2015-04-27
Two very minor changes

jrfnl:
* Don't pass parameter to function which doesn't have parameters
* Only print (and escape) the message when there actually is one

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -439,7 +439,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				<?php $plugin_table->prepare_items(); ?>
 
 				<?php
-				if ( isset( $this->message ) && is_string( $this->message ) && '' !== $this->message ) {
+				if ( isset( $this->message ) && is_string( $this->message ) && ! empty( $this->message ) ) {
 					echo wp_kses_post( $this->message );
 				}
 				?>
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -439,7 +439,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				<?php $plugin_table->prepare_items(); ?>
 
 				<?php
-				if ( isset( $this->message ) ) {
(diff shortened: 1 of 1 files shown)


class-tgm-plugin-activation.php
```diff
@@ -439,7 +439,7 @@ public function install_plugins_page() {
 				<?php $plugin_table->prepare_items(); ?>
 
 				<?php
-				if ( isset( $this->message ) ) {
+				if ( isset( $this->message ) && is_string( $this->message ) && '' !== $this->message ) {
```
GaryJones: Per the other discussion, I think the last part here should use `empty()`.

Alternatively, as things stand, the second part is redundant, since the last part is already checking the type is a string.

So, either:

~~~php
if ( isset( $this->message ) && is_string( $this->message ) && ! empty( $this->message ) ) {
~~~

or

~~~php
if ( isset( $this->message ) && '' !== ( $this->message ) ) {
~~~

should be sufficient.

jrfnl: I'll change it to your first option - the second is definitely incorrect as that would pass `new stdClass`, `(int) 5` etc

jrfnl: #wceu public is a bit too diverse for my hard-core PHP defensive coding talk (_The big "Why equal doesn't equal" Quiz_), but maybe we should do with a select group in a hallway track or something....

jrfnl: Oh and I've update the PR for empty

GaryJones: > I'll change it to your first option - the second is definitely incorrect as that would pass new stdClass, (int) 5 etc

Yes, you're right.

> #wceu public is a bit too diverse for my hard-core PHP defensive coding talk (The big "Why equal doesn't equal" Quiz), but maybe we should do with a select group in a hallway track or something....

I've seen a video of it :-)

jrfnl: Interesting... I always ask them to cut the actual quiz out if they video it - which one did you see ? There are actually some five different versions of the quiz (different sets of questions).

PR #362, opened 2015-04-27
Simplify sorting of table data

jrfnl:


GaryJones:
Could this sorting functionality be pulled out into a new function, so it could potentially be unit tested?

Current function to gather the data, another to sort it in a specific way. Might then allow others to filter the callback so that they could sort it in another way.

jrfnl:
Good point: done!

class-tgm-plugin-activation.php
```diff
@@ -1378,33 +1380,32 @@ protected function _gather_plugin_data() {
...
+		 *
+		 * @return array
+		 */
+		public function sort_table_items( $items ) {
+
+			$type = array();
+			$name = array();
+
+			foreach ( $items as $i => $plugin ) {
+				$type[ $i ] = $plugin['type'];
```
GaryJones: Where is this 'type' value originally being added from? I don't see it in example.php, and we don't populate it in `register()`.

GaryJones:
Certainly better than what we have, but something is niggling me about it.

It orders by "Required" then "Recommended" (descending type), but in other languages these words may be in the reverse order.

Also, it doesn't account for any third option that we may want to add later.

Can there be something like:

~~~php
// Main class
/**
 * Possible types.
 *
 * Ordered by the order they should appear in the plugin table (i.e. Required top by default).
 */
public function get_types() {
    return apply_filters(
        'tgmpa_types',
        array(
            'required'    => __( 'Required', 'tgmpa' ),
            'recommended' => __( 'Recommended', 'tgmpa' ),
        )
    );
}
~~~

And then grab and use this array to check the key and multisort accordingly?

That allows authors to change the order via filter, change all the wording (if we also pull from this method when creating table rows) without filtering gettext or translating, and allows us an easy way to add new types in the future.

jrfnl:
While I agree with your line of thinking and had been wondering about the i18n aspect as well, I think that is a change which should be in another pull request.
This one really is about simplifying the sorting code (of which most was redundant, never used as it was wrongly set up).

Your point is about abstracting advise types and we need to put some more thought into that as I believe we should also abstract the logic for determining the type (sort of duplicated in some places) and if we abstract it, think about things like a 'Buy' type where there would be no install link.
And to be fair, if we take that into account, it sounds more like a 3.0 feature to me.

PR #367, opened 2015-04-28
Code style

jrfnl:
- if/else spacing
- `empty()` rather than `!== ''`
- break down some overly long lines
- minor improvements to a few variable checks

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -219,7 +219,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 * @see TGM_Plugin_Activation::styles()
 		 */
 		public function init() {
-			
+
 			if ( apply_filters( 'tgmpa_load', ! is_admin() ) ) {
 				return;
 			}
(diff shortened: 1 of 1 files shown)


class-tgm-plugin-activation.php
```diff
@@ -926,7 +924,7 @@ public function dismiss() {
 		 * @param array $plugin Array of plugin arguments.
 		 */
 		public function register( $plugin ) {
-			if ( ! isset( $plugin['slug'] ) || ! isset( $plugin['name'] ) ) {
+			if ( empty( $plugin['slug'] ) || empty( $plugin['name'] ) ) {
```
GaryJones: Can this actually be a straight swap, or does isset check if the key exists, and not that it exists with an empty value?

jrfnl: `isset()` checks that the variable exists and is not `null`, `empty()` tests whether the variable exists and is not empty for a variety of definitions of "empty". `empty()` will return `true` for non-existent variables without an undefined variable notice (which is why it's often abused by people who don't understand `empty()` properly).

So in this case, I'm enhancing the variable check by testing for existence and a non-empty value.

GaryJones: Excellent - I've actually taken notice that `empty()` does the exists check or not.

PR #368, opened 2015-04-28
Feature/simplify passed info

jrfnl:
Tested reasonably extensively

### Improve plugin registration method

- register plugins with the `$slug` as associative array key
- ensure that all plugins have defaults registered for the keys not passed
- add `$has_forced_activation` and `$has_forced_deactivation` properties to the TGMPA class which are set from the plugin registration method and used to determine whether to add the relevant actions
- add `$sort_order` property which holds the plugin names for use in sorting the `$plugins` property for internal use
- as - no matter what - we'll always need the `file_path`, we might as well also set it in the registration method and get rid of the repeated calls to `$this->populate_file_path()`
- also made the `populate_file_path()` a little bit more flexible so it can be used in a leaner manner for those cases where it's still needed

### Simplify the table generation, install and activate code

- retrieve most information we need from our `$plugins` array rather than pass it through URL or `$_POST`
- properly sanitize and validate received information from URL or `$_POST`
- leverage the use of the WP_filesystem API
- show an appropriate error message if user tries to activate an already active plugin
- deprecate the `TGMPA_List_Table::_get_plugin_data_from_name()` method which is no longer needed as we have access to the `TGM_Plugin_Activation::_get_plugin_data_from_name()` method
- Fix unauthorized issue on bulk installs when not using direct file access caused by #227
- Only prepare some variables and update recently activated plugins option when we actually need to

Also:
- abstract the regexes which are used a number of times to class constants
- combine multiple `isset()`'s if used within the same if statement
- don't nest what doesn't need nesting
- unset loop variables if the function doesn't return straight after

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -49,7 +49,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 	 */
 	class TGM_Plugin_Activation {
 
-		const WP_REPO_REGEX = '|^http[s]?://wordpress\.org/extend/plugins/|';
+		const WP_REPO_REGEX = '|^http[s]?://wordpress\.org/(?:extend/)?plugins/|';
 
 		const EXT_REPO_REGEX = '|^http[s]?://|';
 
@@ -506,7 +506,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 
 			// All plugin information will be stored in an array for processing.
 			$plugin = array();
-			$slug   = sanitize_title( $_GET['plugin'] );
+			$slug   = sanitize_key( $_GET['plugin'] );
 
 			// Checks for actions from hover links to process the installation.
 			if ( isset( $this->plugins[ $slug ], $_GET['tgmpa-install'] ) && 'install-plugin' === $_GET['tgmpa-install'] ) {
@@ -745,7 +745,6 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			}
 
 			$installed_plugins   = get_plugins(); // Retrieve a list of all the plugins
-
 			$message             = array(); // Store the messages in an array to be outputted after plugins have looped through.
 			$install_link        = false;   // Set to false, change to true in loop if conditions exist, used for action link 'install'.
 			$install_link_count  = 0;       // Used to determine plurality of install action link text.
@@ -953,20 +952,28 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			}
 
 			$defaults = array(
-				'name'               => '',
-				'slug'               => '',
-				'source'             => 'repo',
-				'required'           => false,
-				'version'            => '',
-				'force_activation'   => false,
-				'force_deactivation' => false,
-				'external_url'       => '',
-				'is_callable'        => '',
+				'name'               => '',      // string
+				'slug'               => '',      // string
+				'source'             => 'repo',  // string
+				'required'           => false,   // boolean
+				'version'            => '',      // string
+				'force_activation'   => false,   // boolean
+				'force_deactivation' => false,   // boolean
+				'external_url'       => '',      // string
+				'is_callable'        => '',      // string or array
 			);
 
+			// Prepare the received data
 			$plugin                              = wp_parse_args( $plugin, $defaults );
 			$plugin['file_path']                 = $this->_get_plugin_basename_from_slug( $plugin['slug'] );
 
+			// Forgive users for using string versions of booleans or floats for version nr
+			$plugin['version']                   = (string) $plugin['version'];
+			$plugin['required']                  = TGM_Utils::validate_bool( $plugin['required'] );
+			$plugin['force_activation']          = TGM_Utils::validate_bool( $plugin['force_activation'] );
+			$plugin['force_deactivation']        = TGM_Utils::validate_bool( $plugin['force_deactivation'] );
+
+			// Set the class properties
 			$this->plugins[ $plugin['slug'] ]    = $plugin;
 			$this->sort_order[ $plugin['slug'] ] = $plugin['name'];
 
@@ -1609,7 +1616,7 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
 
 				$plugins_to_install = array();
 
-				if ( isset( $_POST['plugin'] ) && ! empty( $_POST['plugin'] ) ) {
+				if ( ! empty( $_POST['plugin'] ) ) {
 
 					if ( is_array( $_POST['plugin'] ) ) {
 						$plugins_to_install = (array) $_POST['plugin'];
@@ -1618,9 +1625,9 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
 						// Received via Filesystem page - unflatten array (WP bug #19643)
 						$plugins_to_install = explode( ',', $_POST['plugin'] );
 					}
-					
+
 					// Sanitize the received input
-					$plugins_to_install = array_map( 'sanitize_title', $plugins_to_install );
+					$plugins_to_install = array_map( 'sanitize_key', $plugins_to_install );
 
 					// Validate the received input
 					foreach ( $plugins_to_install as $key => $slug ) {
@@ -1737,7 +1744,11 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
 				check_admin_referer( 'bulk-' . $this->_args['plural'] );
 
 				// Grab plugin data from $_POST.
-				$plugins             = isset( $_POST['plugin'] ) ? (array) $_POST['plugin'] : array();
+				$plugins = array();
+				if ( isset( $_POST['plugin'] ) ) {
+					$plugins = array_map( 'sanitize_key', (array) $_POST['plugin'] );
+				}
+
 				$plugins_to_activate = array();
 				$plugin_names        = array();
 
@@ -1820,7 +1831,6 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
 		 *
 		 * @since      2.2.0
 		 * @deprecated 2.5.0
-		 * @deprecated use TGM_Plugin_Activation::_get_plugin_data_from_name()
 		 * @see        TGM_Plugin_Activation::_get_plugin_data_from_name()
 		 *
 		 * @param string $name Name of the plugin, as it was registered.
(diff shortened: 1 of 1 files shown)


class-tgm-plugin-activation.php
```diff
@@ -758,17 +775,17 @@ public function notices() {
 					$install_link = true; // We need to display the 'install' action link.
 					$install_link_count++; // Increment the install link count.
 					if ( current_user_can( 'install_plugins' ) ) {
-						if ( isset( $plugin['required'] ) && $plugin['required'] ) {
-							$message['notice_can_install_required'][] = $plugin['name'];
+						if ( true === $plugin['required'] ) {
```
GaryJones: How about if the user has registered slightly incorrect as a string:

`required => 'true',`

? We should allow a little leeway for 2.5, and then make it stricter for 3.0.

class-tgm-plugin-activation.php
```diff
@@ -726,28 +752,19 @@ public function notices() {
 			$activate_link       = false;   // Set to false, change to true in loop if conditions exist, used for action link 'activate'.
 			$activate_link_count = 0;       // Used to determine plurality of activate action link text.
 
-			foreach ( $this->plugins as $plugin ) {
+			foreach ( $this->plugins as $slug => $plugin ) {
```
GaryJones: Can this section of deciding which messages need showing be refactored out?

class-tgm-plugin-activation.php
```diff
@@ -802,25 +820,24 @@ public function notices() {
 				}
 
 				// Grab all plugin names.
-				foreach ( $message as $type => $plugin_groups ) {
-					$linked_plugin_groups = array();
-
-					// Count number of plugins in each message group to calculate singular/plural message.
-					$count = count( $plugin_groups );
+				foreach ( $message as $type => $plugin_group ) {
```
GaryJones: Can this section of looping through messages and showing them be refactored out?

class-tgm-plugin-activation.php
```diff
@@ -828,22 +845,25 @@ public function notices() {
...
+							$linked_plugins[] = esc_html( $this->plugins[ $plugin_slug ]['name'] ); // No hyperlink.
 						}
 					}
+					unset( $plugin_slug, $url );
 
-					$last_plugin = array_pop( $plugin_groups ); // Pop off last name to prep for readability.
-					$imploded    = empty( $plugin_groups ) ? '<em>' . $last_plugin . '</em>' : '<em>' . ( implode( ', ', $plugin_groups ) . '</em> and <em>' . $last_plugin . '</em>' );
+					$count       = count( $plugin_group );
+					$last_plugin = array_pop( $linked_plugins ); // Pop off last name to prep for readability.
+					$imploded    = empty( $linked_plugins ) ? '<em>' . $last_plugin . '</em>' : '<em>' . ( implode( ', ', $linked_plugins ) . '</em> ' . esc_html__( 'and', 'tgmpa' ) . ' <em>' . $last_plugin . '</em>' );
```
GaryJones: How about a small private/protected function that returns a string in `<em>`??

The comma between each plugin will be italicised, so any semantics the emphasis element is meant to have per plugin, is not applicable.

The `esc_html__()` should be `esc_html_x()` so that some context can be given to translators about what that "and" is for.

class-tgm-plugin-activation.php
jrfnl: Should be done in a separate PR, is a different issue. (and I'm preparing a number of those already, mainly focussed on abstracting (nearly) duplicate code to functions)

class-tgm-plugin-activation.php
jrfnl: Should be done in a separate PR, is a different issue. (and I'm preparing a number of those already, mainly focussed on abstracting (nearly) duplicate code to functions)

class-tgm-plugin-activation.php
```diff
@@ -1029,6 +1080,10 @@ public function populate_file_path() {
 		 */
 		protected function _get_plugin_basename_from_slug( $slug ) {
 
+			if ( ! function_exists( 'get_plugins' ) ) {
+				require_once ABSPATH . 'wp-admin/includes/plugin.php';
```
GaryJones: Is there a `WP_ADMIN` constant?

class-tgm-plugin-activation.php
jrfnl: I've added some type casting to the registration method now to account for this. Using some of the filter/filter emulate code I pointed to elsewhere from the WPSEO plugin. Not giving credit to them as I wrote it ;-)

class-tgm-plugin-activation.php
```diff
@@ -1286,25 +1337,25 @@ protected function _gather_plugin_data() {
...
-
-				$external_url = $this->_get_plugin_data_from_name( $plugin['name'], 'external_url' );
-				$source       = $this->_get_plugin_data_from_name( $plugin['name'], 'source' );
+				$table_data[ $i ]['slug']             = $slug;
 
-				if ( $external_url && preg_match( '|^http(s)?://|', $external_url ) ) {
-					$table_data[ $i ]['plugin'] = '<strong><a href="' . esc_url( $external_url ) . '" target="_blank">' . esc_html( $plugin['name'] ) . '</a></strong>';
-				}
-				elseif ( ! $source || preg_match( '|^http://wordpress.org/extend/plugins/|', $source ) ) {
+				if ( ! empty( $plugin['external_url'] ) && preg_match( TGM_Plugin_Activation::EXT_REPO_REGEX, $plugin['external_url'] ) ) {
```
GaryJones: Since this condition, and the elseif condition below are already being used around line 829, can you create public boolean methods in the main class, and have both conditions here reference each method?
i.e. (as a quick example)
~~~php
// TGM_Plugin_Activation class
public function is_valid_external_source( $external_url ) {
    return ! empty( $external_url ) && preg_match( self::EXT_REPO_REGEX, $external_url );
}
~~~

~~~php
// Used here as:
if ( $this->tgmpa->is_valid_external_source( $plugin['external_url'] ) {
~~~

The elseif below might be `is_valid_repo_source()`, and the if / elseif in the main plugin (line ~829) would be called as `if ( $this->is_valid_external_source( $this->plugins[ $plugin_slug ]['external_url'] )` etc.

class-tgm-plugin-activation.php
```diff
@@ -1286,25 +1337,25 @@ protected function _gather_plugin_data() {
...
+					$table_data[ $i ]['plugin'] = sprintf(
+						'<strong><a href="%1$s" target="_blank">%2$s</a></strong>',
+						esc_url( $plugin['external_url'] ),
+						esc_html( $plugin['name'] )
+					);
+				} elseif ( 'repo' === $plugin['source'] || preg_match( TGM_Plugin_Activation::WP_REPO_REGEX, $plugin['source'] ) ) {
 					$url = add_query_arg(
```
GaryJones: This `$url` generation could be refactored  with the earlier one too.

class-tgm-plugin-activation.php
jrfnl: I like idea, but again, should be done in a separate PR, not the same issue.

class-tgm-plugin-activation.php
```diff
@@ -1862,6 +1816,26 @@ public function prepare_items() {
...
 
+
+		/* *********** DEPRECATED METHODS *********** */
+
+		/**
+		 * Retrieve plugin data, given the plugin name.
+		 *
+		 * @since      2.2.0
+		 * @deprecated 2.5.0
+		 * @deprecated use TGM_Plugin_Activation::_get_plugin_data_from_name()
```
GaryJones: This line is redundant.

class-tgm-plugin-activation.php
```diff
@@ -1501,18 +1528,12 @@ public function column_plugin( $item ) {
...
-			if ( __( 'Private Repository', 'tgmpa' ) === $item['source'] ) {
-				// Escape external URLs
-				$plugin_url = esc_url( $plugin_url );
-			}
-			elseif ( __( 'Pre-Packaged', 'tgmpa' ) === $item['source'] ) {
-				// Encode file path for use in attribute
-				$plugin_url = urlencode( $this->tgmpa->default_path . $plugin_url );
-			}
-
-			$value = $item['file_path'] . ',' . $plugin_url . ',' . $item['sanitized_plugin'];
```
GaryJones: Was this just a hacky way to get all of the information into `$_POST`?

class-tgm-plugin-activation.php
jrfnl: There is, but it's a boolean, so not a path to admin ;-)

class-tgm-plugin-activation.php
```diff
@@ -1587,176 +1607,113 @@ public function process_bulk_actions() {
...
-				}
-				// Looks like the user can use the direct method, take from $_POST.
-				elseif ( isset( $_POST['plugin'] ) ) {
-					$plugins = (array) $_POST['plugin'];
-				}
-				// Nothing has been submitted.
-				else {
-					$plugins = array();
-				}
+				if ( isset( $_POST['plugin'] ) && ! empty( $_POST['plugin'] ) ) {
```
GaryJones: I thought you said `empty()` already checks for `isset()`?

class-tgm-plugin-activation.php
jrfnl: https://wordpress.org/plugins/debug-bar-constants/ FTW :sunglasses:

class-tgm-plugin-activation.php
```diff
@@ -1587,176 +1607,113 @@ public function process_bulk_actions() {
...
-							$plugins_to_install[] = $value;
-							$plugin_installs[]    = $value;
+					
+					// Sanitize the received input
+					$plugins_to_install = array_map( 'sanitize_title', $plugins_to_install );
+
+					// Validate the received input
+					foreach ( $plugins_to_install as $key => $slug ) {
+						// Check if the plugin was registered with TGMPA and remove if not
+						if ( ! isset( $this->tgmpa->plugins[ $slug ] ) ) {
```
GaryJones: Is this logic correct?

"If NOT a registered plugin, ignore this one we want to install"?

class-tgm-plugin-activation.php
jrfnl: I'm abstracting this (whole block of code) in a different way in one of the new branches

class-tgm-plugin-activation.php
jrfnl: Dito

class-tgm-plugin-activation.php
jrfnl: Removed

class-tgm-plugin-activation.php
jrfnl: Hell yeah

class-tgm-plugin-activation.php
jrfnl: Good catch

class-tgm-plugin-activation.php
```diff
@@ -1587,176 +1607,113 @@ public function process_bulk_actions() {
...
-				// Loop through each plugin to install and try to grab information from WordPress API, if not create 'tgmpa-empty' scalar.
-				$i = 0;
-				foreach ( $plugin_installs as $plugin ) {
-					$temp = plugins_api( 'plugin_information', array( 'slug' => $plugin, 'fields' => array( 'sections' => false ) ) );
+				// Prepare the data for validated plugins for the install
+				foreach ( $plugins_to_install as $slug ) {
+					$name   = $this->tgmpa->plugins[ $slug ]['name'];
+					$source = $this->tgmpa->plugins[ $slug ]['source'];
+
+					if ( 'repo' === $source || preg_match( TGM_Plugin_Activation::WP_REPO_REGEX, $source ) ) {
```
GaryJones: Here's another instance where the earlier mentioned refactoring could be used.

class-tgm-plugin-activation.php
jrfnl: Logic is correct. We only want to deal with plugins which were registered in TGMPA, if a plugin was not registered by a theme or plugin using TGMPA and still somehow managed to get into the $_POST array, we ignore it as it's probably some sort of hack.

class-tgm-plugin-activation.php
jrfnl: See above

GaryJones:
Phew! Excellent stuff. I realise some of my suggestions may well be other PRs once this is merged, but wanted to mention them to get them on your radar if they weren't already.

PR #373, opened 2015-04-28
Abstract duplicate code to retrieve a download url

jrfnl:


class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1108,75 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
...
+		 * Retrieve the download url for a package.
+		 *
+		 * @since 2.5.0
+		 *
+		 * @param string $slug
+		 *
+		 * @return string Plugin download URL or path to local file
+		 */
+		public function get_download_url( $slug ) {
+			$dl = '';
```
GaryJones: Scrutinizer will likely pick up on this as being too short a variable name. `$download_url` or `$url` would be a little more descriptive.

class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1108,75 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
...
+
+			if ( ! empty( $this->plugins[ $slug ]['source'] ) ) {
```
GaryJones: I prefer to see big conditions like this returned early as a guard clause.

~~~php
if ( empty( $this->plugins[ $slug ]['source'] ) ) {
    return '';
}
~~~

Use of `''` makes it clearer that we're intentionally returning an empty string, and not a valid URL.

class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1108,75 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
 		}
 
 		/**
+		 * Retrieve the download url for a package.
+		 *
+		 * @since 2.5.0
+		 *
+		 * @param string $slug
+		 *
+		 * @return string Plugin download URL or path to local file
```
GaryJones: It's possible for an empty string to be returned when the source is empty, but the docs don't reflect this.

class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1108,75 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
...
+		 */
+		public function get_download_url( $slug ) {
+			$dl = '';
+
+			if ( ! empty( $this->plugins[ $slug ]['source'] ) ) {
+
+				$source = $this->plugins[ $slug ]['source'];
+
+				// Is this a WP repo plugin ?
+				if ( 'repo' === $source || preg_match( self::WP_REPO_REGEX, $source ) ) {
```
GaryJones: Is this is now the only usage of the class constants? If so, may be cleaner to just have them as variables in this function - no point in making stuff more public than needs to be.

class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1108,75 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
...
+
+			// Try to grab information from WordPress API
+			$api = plugins_api( 'plugin_information', array( 'slug' => $slug, 'fields' => array( 'sections' => false ) ) );
+			if ( is_wp_error( $api ) ) {
+				if ( true === WP_DEBUG ) {
+					wp_die( esc_html( $this->strings['oops'] ) . var_dump( $api ) ); // wpcs: xss ok
+				} else {
+					wp_die( esc_html( $this->strings['oops'] ) );
+				}
+			}
```
GaryJones: Code standards :-)

class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1108,75 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
...
+			if ( ! function_exists( 'plugins_api' ) ) {
+				require_once ABSPATH . 'wp-admin/includes/plugin-install.php'; // Need for plugins_api.
+			}
+
+			$source = '';
+
+			// Try to grab information from WordPress API
+			$api = plugins_api( 'plugin_information', array( 'slug' => $slug, 'fields' => array( 'sections' => false ) ) );
+			if ( is_wp_error( $api ) ) {
+				if ( true === WP_DEBUG ) {
```
GaryJones: This has been bugging me.

Cleaner here would be:

~~~php
$dump = ( true === WP_DEBUG ) ? var_dump( $api ) : ''; //wpcs: xss ok
wp_die( esc_html( $this->strings['oops'] ) . $dump );
~~~

class-tgm-plugin-activation.php
GaryJones: I see it's [not](https://github.com/thomasgriffin/TGM-Plugin-Activation/pull/374/files#diff-6f3dff769e44625dd72b031675006eb3R1115), but as before, these conditions could be wrapped in boolean methods so there is only one instance of each.

class-tgm-plugin-activation.php
jrfnl: Hmm.. but $url would be wrong as it can be a path just as easily... which is why I went with $dl ;-)

class-tgm-plugin-activation.php
jrfnl: Having it at the top makes no difference in speed or execution or the function.

class-tgm-plugin-activation.php
jrfnl: Could, true, but IMHO this is already a big step forward. There are other things in the code which are more important to fix than adding those boolean methods.

class-tgm-plugin-activation.php
jrfnl: That will cause havoc for anyone who's got the Xdebug extension running as xdebug will prettify the output using html.

jrfnl:
Updated the PR

class-tgm-plugin-activation.php
GaryJones: Then the function name is misleading too :-)

class-tgm-plugin-activation.php
GaryJones: No, but it unindents the majority of the function, and makes it easier for humans to parse. Even PEAR has it as a best practice: https://pear.php.net/manual/en/standards.bestpractices.php

class-tgm-plugin-activation.php
GaryJones: My code should be no different - just refactoring out multiple calls to wp_die(), esc_html() and array value access.

PR #374, opened 2015-04-28
Abstract duplicate code to retrieve a plugin information link

jrfnl:


class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1101,50 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
...
 		/**
+		 * Retrieve a link to a plugin information page
+		 *
+		 * @since 2.5.0
+		 *
+		 * @param string $slug
+		 *
+		 * @return string
+		 */
+		public function get_plugin_info_link( $slug ) {
```
GaryJones: Why is this one named with "plugin", yet the `get_download_link()` method isn't? Both are talking about plugins, so some consistency either way would be nice.

class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1101,50 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
 		}
 
 		/**
+		 * Retrieve a link to a plugin information page
+		 *
+		 * @since 2.5.0
+		 *
+		 * @param string $slug
+		 *
+		 * @return string
```
GaryJones: A bit of description here would always be helpful.

class-tgm-plugin-activation.php
```diff
@@ -1128,6 +1101,50 @@ public function _get_plugin_data_from_name( $name, $data = 'slug' ) {
...
+		 */
+		public function get_plugin_info_link( $slug ) {
+			$link = '';
+
+			if ( ! empty( $this->plugins[ $slug ]['external_url'] ) && preg_match( self::EXT_REPO_REGEX, $this->plugins[ $slug ]['external_url'] ) ) {
+				$link = sprintf(
+					'<a href="%1$s" target="_blank">%2$s</a>',
+					esc_url( $this->plugins[ $slug ]['external_url'] ),
+					esc_html( $this->plugins[ $slug ]['name'] )
```
GaryJones: Not related to this PR, but as a side note for future consideration.

With all of these "get some value about a single plugin", it might be nice to have the plugins registered as arrays, but stored as a `TGM_Plugin`, so that each value can grabbed as a property (either actual properties, or via `__get()` with whitelist etc.

~~~php
// top of this function
$plugin = $this->plugins[ $slug ]

// This line:
esc_url( $plugin->external_url ),
esc_html( $plugin->name )
~~~

(Yes, the "top of this function" shortcut could be used anyway.)

jrfnl: True, same for abstracting the whole notices functionality to it's own class. There's plenty to do, but not everything needs to be done now.

class-tgm-plugin-activation.php
jrfnl: Because the WP api for this calls it plugin information...

jrfnl:
PR updated

class-tgm-plugin-activation.php
GaryJones: Hence "for future consideration".

jrfnl: You might want to open issues for all the 'for future consideration' improvements. Makes more sense to me then burying them in PRs which will be merged & closed soon.

PR #375, opened 2015-04-28
Abstract single plugin activation to seperate method

jrfnl:


class-tgm-plugin-activation.php
```diff
@@ -588,17 +588,12 @@ protected function do_plugin_install() {
...
-					$this->populate_file_path( $slug ); // Re-populate the file path now that the plugin has been installed and activated.
-
-					if ( is_wp_error( $activate ) ) {
-						echo '<div id="message" class="error"><p>', wp_kses_post( $activate->get_error_message() ), '</p></div>',
-							'<p><a href="', esc_url( add_query_arg( 'page', urlencode( $this->menu ), self_admin_url( $this->parent_slug ) ) ), '" target="_parent">', esc_html( $this->strings['return'] ), '</a></p>';
-						return true; // End it here if there is an error with automatic activation
-
-					} else {
-						echo '<p>', esc_html( $this->strings['plugin_activated'] ), '</p>';
+					if( $this->activate_single_plugin( $plugin_activate, $slug, true ) === true ) {
```
GaryJones: Yoda conditions please.

class-tgm-plugin-activation.php
```diff
@@ -719,6 +693,54 @@ public function maybe_adjust_source_dir( $source, $remote_source, $upgrader ) {
...
+		 * Activate a single plugin and send feedback about the result to the screen.
+		 *
+		 * @since 2.5.0
+		 *
+		 * @param string $file_path Path within wp-plugins/ to main plugin file
+		 * @param string $slug      Plugin slug
+		 * @param bool   $automatic Whether this is an automatic activation after an install. Defaults to false.
+		 *                          This determines the styling of the output messages.
+		 *
+		 * @return bool True if an error was encountered, false otherwise.
```
GaryJones: Returning `true` for a failure, and `false` for success seems back to front.

class-tgm-plugin-activation.php
```diff
@@ -719,6 +693,54 @@ public function maybe_adjust_source_dir( $source, $remote_source, $upgrader ) {
...
+		 */
+		protected function activate_single_plugin( $file_path, $slug, $automatic = false ) {
+
+			if ( ! is_plugin_active( $file_path ) ) {
+				$activate = activate_plugin( $file_path ); // Activate the plugin.
+
+				if ( is_wp_error( $activate ) ) {
+					echo '<div id="message" class="error"><p>', wp_kses_post( $activate->get_error_message() ), '</p></div>',
+						'<p><a href="', esc_url( add_query_arg( 'page', urlencode( $this->menu ), self_admin_url( $this->parent_slug ) ) ), '" target="_parent">', esc_html( $this->strings['return'] ), '</a></p>';
```
GaryJones: This might overlap with a previous PR to get a URL.

class-tgm-plugin-activation.php
```diff
@@ -719,6 +693,54 @@ public function maybe_adjust_source_dir( $source, $remote_source, $upgrader ) {
...
+					}
+					else {
+						// Simpler message layout for use on the plugin install page
+						echo '<p>', esc_html( $this->strings['plugin_activated'] ), '</p>';
+					}
+				}
+			} else {
+				// No simpler message format provided as this message should never be encountered
+				// on the plugin install page
+				echo '<div id="message" class="error"><p>',
```
GaryJones: What if others are also outputting admin notices? Will there be multiple elements with the same ID?

class-tgm-plugin-activation.php
```diff
@@ -719,6 +693,54 @@ public function maybe_adjust_source_dir( $source, $remote_source, $upgrader ) {
...
+				} else {
+					if ( ! $automatic ) {
+						// Make sure message doesn't display again if bulk activation is performed immediately after a single activation.
+						if ( ! isset( $_POST['action'] ) ) {
+							echo '<div id="message" class="updated"><p>', esc_html( $this->strings['activated_successfully'] ), ' <strong>', esc_html( $this->plugins[ $slug ]['name'] ), '.</strong></p></div>';
+						}
+					}
+					else {
+						// Simpler message layout for use on the plugin install page
+						echo '<p>', esc_html( $this->strings['plugin_activated'] ), '</p>';
```
GaryJones: Instead of hard-coding `<p>`, just use `wpautop()`?

(Same for other instances?)

class-tgm-plugin-activation.php
```diff
@@ -744,7 +766,7 @@ public function notices() {
 				return;
 			}
 
-			$installed_plugins   = get_plugins(); // Retrieve a list of all the plugins
+			$installed_plugins = get_plugins(); // Retrieve a list of all the plugins
```
GaryJones: Alignment :-)

class-tgm-plugin-activation.php
jrfnl: Totally, but see the code above the previous comment, I was just keeping in line with the convulsed logic of WP in this case (returning true to stop display of the rest of the page).

class-tgm-plugin-activation.php
jrfnl: These messages are only displayed on our pages and in the case of the ones with the 'message' id - on the actual plugin installation page used by TGMPA. No other messages will (should) display there.

class-tgm-plugin-activation.php
jrfnl: Ouch, no - massive overhead which we can do without
See: https://core.trac.wordpress.org/browser/tags/4.2.1/src/wp-includes/formatting.php#L373

jrfnl:
Updated & rebased

PR #376, opened 2015-04-28
Add some html wrapper methods for use with array_map() ...

jrfnl:
...and implement their use.

Also:
Fix some html errors in the notice output

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -796,11 +796,15 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			// If we have notices to display, we move forward.
 			if ( ! empty( $message ) ) {
 				krsort( $message ); // Sort messages.
-				$rendered = ''; // Display all nag messages as strings.
+				$rendered = '';
+
+				// As add_settings_error() wraps the final message in a <p> and as the final message can't be
+				// filtered, using <p>'s in our html would render invalid html output.
+				$line_template = '<span style="display: block; margin: 0.5em 0.5em 0 0; clear: both;">%s</span>' . "\n";
 
 				// If dismissable is false and a message is set, output it now.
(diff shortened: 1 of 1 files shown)


class-tgm-plugin-activation.php
```diff
@@ -814,11 +818,19 @@ public function notices() {
...
-					$imploded    = empty( $linked_plugins ) ? '<em>' . $last_plugin . '</em>' : '<em>' . ( implode( ', ', $linked_plugins ) . '</em> ' . esc_html__( 'and', 'tgmpa' ) . ' <em>' . $last_plugin . '</em>' );
+					$count          = count( $plugin_group );
+					$linked_plugins = array_map( array( 'TGM_Utils', 'wrap_in_em' ), $linked_plugins );
+					$last_plugin    = array_pop( $linked_plugins ); // Pop off last name to prep for readability.
+					$imploded       = empty( $linked_plugins ) ? $last_plugin : ( implode( ', ', $linked_plugins ) . ' ' . esc_html_x( 'and', 'plugin A *and* plugin B', 'tgmpa' ) . ' ' . $last_plugin );
 
-					$rendered .= '<p>' . sprintf( translate_nooped_plural( $this->strings[ $type ], $count, 'tgmpa' ), $imploded, $count ) . '</p>';
+					$rendered .= sprintf(
+						$line_template,
+						sprintf(
```
GaryJones: Do we need any checks here to make sure both plural and singular versions of `$this->strings[ $type ]` have exactly two placeholders?

jrfnl: If we'd go that way, we'd have to do that for quite a few strings I think - would be more appropriate in a method which reads in all the strings. Then again - what to do if a custom string or a translation doesn't have the right nr of placeholders ? Fall back to our default in English ? That could make for some strange mixed language pages, including mixed terminology (plugin/theme requires .. etc).
Let leave it to the configurators and translators for now until someone reports an issue with it.

PR #378, opened 2015-04-29
Feature/housekeeping

jrfnl:
See individual commits. Mostly code readability (break down excessively long lines) and removal of use of `compact()`.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -50,12 +50,12 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 	class TGM_Plugin_Activation {
 
 		/**
-		 * @const string Regular expression to test if a url is a WP plugin repo url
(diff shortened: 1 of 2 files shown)


jrfnl:
Shall I send in a PR with the sniffs I've used now ?

P.S.: re WPCS#241 - the Yoast CS which Rarst refers to might be useful to look at - I worked on those as well. 
Secondly, that discussion mentions parsing of hook documentation. Here's a (working) proof of concept I did for that some time ago: https://github.com/jrfnl/wp-hook-documentor

GaryJones:
> Shall I send in a PR with the sniffs I've used now ?

To WPCS or TGMPA?

jrfnl:
Either - to TGMPA at first so we can do some fine-tuning and if it feels right, to WPCS ?

GaryJones:
Sure :+1:

PR #351, opened 2015-04-26
Update Feature (was Develop)

swiderski:
Update feature! 
Hopefully this time I will do it right.

PR #350, opened 2015-04-26
Add update mode

swiderski:
I am n00b, too much working alon... but I removed files of IDE, added gitignore, and merged with current develop version. Hope this time will be OK.

PR #381, opened 2015-05-03
Add updating workflow

### LINKED ISSUE #192: Upgrading the package plugins
Hi Thomas

I am having a problem with the TGM plugin when I am trying to upgrade plugins which I have already installed using the TGM plugin.  When you change the version number in the configuration, the TGM plugin correctly displays that the plugin needs to be upgraded and displays "The following plugin needs to be updated to its latest version to ensure maximum compatibility with this theme".  The problem is the "Install Plugins" page is never re-activated and I do not get an option to update the plugin as shown in this video http://www.screenr.com/uYmH.  Is there a setting which I am missing apart from updating the version number?  Or do I have to re-named the packaged .zip file?  

At the moment the workaround is to deactivated the theme then uninstall the plugin.  Then, re-activate the theme and then the TGM plugin will install the new plugin.  This is not ideal as it does not display a maintenance mode and also it changes the theme on the front end.

This is an awesome solution overall and I must just be missing something.  I can see that the solution I am looking for is on this TGM plugin video url above.  But im wondering if you have changed the feature since this was made.

Thank you very much.

jrfnl:
Closes #192, #350, #351 

The commits have been set up to facilitate easier code review as these are quite significant changes. Please see the individual commit messages for detailed information on what I've changed, how and why.

#### Functional highlights:
* Updating can now be done from the TGMPA screen, both on individual plugins as well as in bulk - this will take into account WP repo updates as well as updates for plugins which are bundled or come from external sources where a minimum version is set which is higher than the current version.
* Users will be notified of available updates.
* TGMPA screen now has four views: _all_, _to install_, _update available_ and _to activate_
* TGMPA screen now has - on selected views - an extra column showing relevant plugin version information.
* TGMPA screen status column will show both install/activate as well as update status (cumulative).
* If a plugin **_requires_** a certain minimum version of a plugin and the currently installed version does not comply, activation will be blocked until the user has upgraded. If the plugin is already active, it will *not* be deactivated however.
	* If the required plugin version itself requires a higher WP version than the currently installed WP, upgrade will be blocked
* The plugin action links on the WP native plugins page will reflect this too - including disabling deactivation if `force_activation` is `true` for a plugin

Branch has been tested, but further testing is very welcome /cc @duckzland @swiderski
Testing on setups with non-direct filesystem access would be much appreciated (should work, but rather would get it confirmed).

Should work with WP 3.7 and higher.

Props [Zauan/Hogash Studio], [Christian], [Franklin Gitonga], [Jason Xie], [swiderski] for the inspiration.

#### Screenshots:
screenshot
screenshot

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -850,50 +850,32 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 */
 		public function notices() {
 
-			// Remove nag on the install page.
-			if ( $this->is_tgmpa_page() ) {
+			// Remove nag on the install page / Return early if the nag message has been dismissed.
+			if ( $this->is_tgmpa_page() || get_user_meta( get_current_user_id(), 'tgmpa_dismissed_notice_' . $this->id, true ) ) {
 				return;
 			}
 
-			// Return early if the nag message has been dismissed.
-			if ( get_user_meta( get_current_user_id(), 'tgmpa_dismissed_notice_' . $this->id, true ) ) {
-				return;
-			}
+			// Store for the plugin slugs by message type.
+			$message = array();
 
-			$installed_plugins   = get_plugins(); // Retrieve a list of all the plugins
-			$message             = array(); // Store the messages in an array to be outputted after plugins have looped through.
-			$install_link        = false;   // Set to false, change to true in loop if conditions exist, used for action link 'install'.
-			$install_link_count  = 0;       // Used to determine plurality of install action link text.
-			$activate_link       = false;   // Set to false, change to true in loop if conditions exist, used for action link 'activate'.
-			$activate_link_count = 0;       // Used to determine plurality of activate action link text.
+			// Initialize counters used to determine plurality of action link texts
+			$install_link_count  = 0;
+			$update_link_count   = 0;
+			$activate_link_count = 0;
 
 			foreach ( $this->plugins as $slug => $plugin ) {
-				// If the plugin is installed and active, check for minimum version argument before moving forward.
-				if ( is_plugin_active( $plugin['file_path'] ) || ( ! empty( $plugin['is_callable'] ) && is_callable( $plugin['is_callable'] ) ) ) {
 
-					// If the current version is less than the minimum required version, we display a message.
-					if ( version_compare( $installed_plugins[ $plugin['file_path'] ]['Version'], $plugin['version'], '<' ) ) {
-						if ( current_user_can( 'install_plugins' ) ) {
-							$message['notice_ask_to_update'][] = $slug;
-						} else {
-							$message['notice_cannot_update'][] = $slug;
-						}
-					}
-					// No minimum version specified or can't find the plugin, so iterate over the plugin.
-					else {
-						continue;
-					}
+				if ( $this->is_plugin_active( $slug ) && false === $this->does_plugin_have_update( $slug ) ) {
+					continue;
 				}
 
-				// Not installed.
-				if ( ! isset( $installed_plugins[ $plugin['file_path'] ] ) ) {
-					$install_link = true; // We need to display the 'install' action link.
-					$install_link_count++; // Increment the install link count.
+				if ( ! $this->is_plugin_installed( $slug ) ) {
+					$install_link_count++;
+
 					if ( current_user_can( 'install_plugins' ) ) {
 						if ( true === $plugin['required'] ) {
 							$message['notice_can_install_required'][] = $slug;
 						}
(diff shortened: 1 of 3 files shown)


KingYes:
+1 good job

duckzland:
@jrfnl got a bug : 
click begin updating -> click return to plugin installer -> do bulk install submit -> click return to dashboard = you got "You do not have sufficient permissions to access this page."

GaryJones:
General comment - do all of the new methods need to be public? Could some be protected?

jrfnl:
> General comment - do all of the new methods need to be public? Could some be protected?

Unfortunately yes. For the TGMPA class methods - nearly all of them are also called from the List_Table/Bulk_Installer/Bulk_Skin classes and therefore those need to be public.

For the new methods in the List_Table class - the ones overloading methods in the WP native class have to be public for the code to be compatible with WP < 4.x The scope keywords were left off in the WP versions before that time, so you'd get a E_STRICT warning about the scope being wrong if they are not public.

nomansbeauty:
great work guys, we will test it out soon. Thanks :+1:

jrfnl:
> @jrfnl got a bug :
click begin updating -> click return to plugin installer -> do bulk install submit -> click return to dashboard = you got "You do not have sufficient permissions to access this page."

@duckzland I've looked into this and believe I've been able to reproduce the issue and fix it. Could you please test again ?

jrfnl:
@GaryJones Thanks for reviewing. I've made the necessary adjustments & pushed.

duckzland:
@jrfnl I've tested it and seems its working as intended, I think it is better to stop testing untill the code is ready for testing as now it seems that too much changes going on.

jrfnl:
@duckzland Thanks for testing.

What do you mean by :

> I think it is better to stop testing untill the code is ready for testing as now it seems that too much changes going on.

Since this PR was merged beginning of the week, nothing has changed which should impact users.

duckzland:
@jrfnl so it is ready for testing with no more major changes and just for bug hunting for 2.5.0 release at this stage?

jrfnl:
From my point of view, I don't expect any more big changes for the 2.5.0 release, so yes, testing is very welcome.

steveboj:
I found an issue with packaged plugins that have uppercase letters in their slug, e.g. 'LayerSlider'. and 'Ultimate_VC_Addons'. The install and update options both failed with no errors for these plugins, but worked perfectly when I converted them to lowercase.

Everything else I tested worked very well.

BassemN:
@steveboj I have the same issue

jrfnl:
@steveboj @BassemN Is this only when updating or also when first installing the package from a bundled plugin ? If the latter - please open a separate issue.

steveboj:
@jrfnl This is happening on first install and when updating. I have just created a separate issue https://github.com/TGMPA/TGM-Plugin-Activation/issues/401

steveboj:
I'm having trouble getting the new update function working with multiple bundled plugins from zip files:

Scenario 1: I click the "update" link for an individual plugin and I receive the "The plugin is at the latest version." message. 

Scenario 2: I click the "update" link for a plugin and I see the success messages, but when I return to the updater page it says that plugin still needs to be updated. I click the "update" link again, but this time it shows me the "The plugin is at the latest version." message. When I check the plugins folder it still contains the old copy of the plugin.

Scenario 3: I use the bulk "update" function and it goes through each plugin saying they have been updated successfully, but when I return to the plugins page it shows me none of the plugins have been updated. When I try to update any of them individually, I get the "The plugin is at the latest version." message.

Scenario 4: I click the "update" link for an individual plugin in the list and it updates perfectly.

Has anyone else experienced anything similar?

jrfnl:
@steveboj Updating with bundled plugins can only be done by setting a minimum required version of the plugin in the settings file - the `version` array key. Set this to the new version bundled and things should be fine.

TGMPA cannot obtain information about the version nr of bundled plugins, so we need to rely on the information you provide in the settings file to determine if there is an update.

steveboj:
@jrfnl The versions numbers seem to all be correctly setup. I'll test some more and see if I can find any more clues. Let me know if you think of anything else that could cause this type of issue.

jrfnl:
@steveboj Great - any more clues you can give me to track down why it's not working for you, would be helpful. Insight into your settings and possibly copies of the offending .zip files could help too (old version + new version).
If the issue persists - please open a separate issue for this to keep it on the radar (this issue is closed).

nomansbeauty:
Seems like new changes works well. Any idea, when the new version or stable release is coming? To be used in production.

GaryJones:
When it's ready - hopefully, before the end of the month.

PR #384, opened 2015-05-04
Housekeeping WPCS / Docs 

jrfnl:
Solve nearly all issues which would start to pop up once WordPress-Doc sniffs are included.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -502,7 +502,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 * @global string $tab Used as iframe div class names, helps with styling
 		 * @global string $body_id Used as the iframe body ID, helps with styling
 		 *
-		 * @return void
+		 * @return null Returns early if not the TGMPA page.
 		 */
 		public function admin_init() {
 
@@ -607,7 +607,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 *
 		 * @since 1.0.0
 		 *
-		 * @return void
+		 * @return null Aborts early if we're processing a plugin installation action.
 		 */
 		public function install_plugins_page() {
 
@@ -646,8 +646,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		}
 
 		/**
-		 * Installs a plugin or activates a plugin depending on the hover
-		 * link clicked by the user.
+		 * Installs, updates or activates a plugin depending on the action link clicked by the user.
 		 *
 		 * Checks the $_GET variable to see which actions have been
 		 * passed and responds with the appropriate method.
@@ -680,7 +679,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				return false;
 			}
 
-			// Checks for actions from hover links to process the installation.
+			// Checks for actions from action links to process the installation.
 			if ( ( isset( $_GET['tgmpa-install'] ) && 'install-plugin' === $_GET['tgmpa-install'] ) || ( isset( $_GET['tgmpa-update'] ) && 'update-plugin' === $_GET['tgmpa-update'] ) ) {
 
 				$install_type = 'install';
@@ -796,7 +795,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				return true;
 			}
 
-			// Checks for actions from hover links to process the activation.
+			// Checks for actions from link-clicks to process the activation.
 			elseif ( isset( $this->plugins[ $slug ]['file_path'], $_GET['tgmpa-activate'] ) && 'activate-plugin' === $_GET['tgmpa-activate'] ) {
 				check_admin_referer( 'tgmpa-activate', 'tgmpa-nonce' );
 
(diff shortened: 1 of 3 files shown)


class-tgm-plugin-activation.php
```diff
@@ -10,8 +10,7 @@
  * @package   TGM-Plugin-Activation
  * @version   2.5.0-alpha
  * @link      http://tgmpluginactivation.com/
- * @author    Thomas Griffin
- * @author    Gary Jones
+ * @author    Thomas Griffin, Gary Jones
```
GaryJones: You seem to be missing a name...

class-tgm-plugin-activation.php
```diff
@@ -488,11 +495,14 @@ public function filter_plugin_action_links_update( $actions ) {
...
 		 *
+		 * Returns early if not the TGMPA page.
+		 *
 		 * @since 2.1.0
 		 *
 		 * @global string $tab Used as iframe div class names, helps with styling
 		 * @global string $body_id Used as the iframe body ID, helps with styling
-		 * @return null Returns early if not the TGMPA page.
+		 *
+		 * @return void
```
GaryJones: WP does not use return void. That was clarified last year. If there's no `return`, there's no `@return`.

class-tgm-plugin-activation.php
```diff
@@ -593,10 +603,11 @@ protected function add_admin_menu( array $args ) {
 		 *
 		 * This method is the callback for the admin_menu method function.
 		 * This displays the admin page and form area where the user can select to install and activate the plugin.
+		 * Aborts early if we're processing a plugin installation action.
 		 *
 		 * @since 1.0.0
 		 *
-		 * @return null Aborts early if we're processing a plugin installation action.
```
GaryJones: If there is a simple `return;` then the type returned is null.

class-tgm-plugin-activation.php
```diff
@@ -1219,24 +1235,24 @@ public function register( $plugin ) {
...
-				'is_callable'        => '',      // String or array
+				'is_callable'        => '',      // String|Array
 			);
 
-			// Prepare the received data
+			// Prepare the received data.
 			$plugin = wp_parse_args( $plugin, $defaults );
 
-			// Forgive users for using string versions of booleans or floats for version nr
+			// Forgive users for using string versions of booleans or floats for version nr.
```
GaryJones: "number". No abbreviations needed in this context.

class-tgm-plugin-activation.php
```diff
@@ -1258,7 +1274,8 @@ public function register( $plugin ) {
 		 *
 		 * @param string $source The source of the plugin as provided, either empty (= WP repo), a file path
 		 *                       (= bundled) or an external URL.
-		 * @return string        'repo', 'external', or 'bundled'
+		 *
+		 * @return string 'repo', 'external', or 'bundled'
```
GaryJones: I prefer space between `@param` and `@return`, like this, but unfortunately the WP handbook says not to.

class-tgm-plugin-activation.php
```diff
@@ -2797,17 +2840,18 @@ public function prepare_items() {
...
-		// *********** DEPRECATED METHODS *********** //
+		/* *********** DEPRECATED METHODS *********** */
 
 		/**
 		 * Retrieve plugin data, given the plugin name.
 		 *
 		 * @since      2.2.0
-		 * @deprecated 2.5.0
+		 * @deprecated 2.5.0 use {@see TGM_Plugin_Activation::_get_plugin_data_from_name()} instead.
 		 * @see        TGM_Plugin_Activation::_get_plugin_data_from_name()
```
GaryJones: Don't need this `@see` if there's an inline `{@see}` above.

class-tgm-plugin-activation.php
```diff
@@ -3406,7 +3459,12 @@ public function after_flush_output() {
 	class TGM_Utils {
 
 		/**
-		 * @var bool $has_filters Whether the PHP filter extension is enabled
+		 * Whether the PHP filter extension is enabled.
+		 *
+		 * @var bool $has_filters True is the extension is enabled.
+		 *
+		 * @see http://php.net/book.filter
```
GaryJones: Like `@param` and `@return` come last, I think `@var` should come last too. The `@see` should come immediately after the description.

Also `@since` missing.

ruleset.xml
```diff
@@ -7,5 +7,6 @@
 	<arg value="sp"/>
 	<rule ref="WordPress">
 		<exclude name="WordPress.VIP" />
+		<exclude name="Squiz.Commenting.ClassComment.TagNotAllowed" />
```
GaryJones: What tags are you using on classes that Squiz says is not allowed?

class-tgm-plugin-activation.php
jrfnl: Actually - the handbook says you do: https://make.wordpress.org/core/handbook/inline-documentation-standards/php-documentation-standards/#1-2-deprecated-functions

ruleset.xml
jrfnl: Nearly everything - including the @since you mention above

class-tgm-plugin-activation.php
jrfnl: Not for me to add ;-)

class-tgm-plugin-activation.php
jrfnl: Then we need to exclude another rule - will add comment to the WPCS thread.

class-tgm-plugin-activation.php
jrfnl: And again, we need to exclude another rule then - well at least in WPCS, I think we should keep it separate like this in TGMPA as we both prefer it ;-)

class-tgm-plugin-activation.php
jrfnl: And yet another exclusion which should be added then.

jrfnl:
Adjusted & pushed + extra comments to WPCS Doc sniff thread added

class-tgm-plugin-activation.php
GaryJones: Always good to learn and brush up on these silly rules :-)

GaryJones:
I had confirmation from @drewapicture regarding `@return null` too, so we're all good there.

jrfnl:
Good - do see my comments on the WPCS-Docs commit then as that does mean that another rule needs to be excluded.

PR #387, opened 2015-05-04
Add a TGMPA version number constant.

jrfnl:
This can/will be useful for theme/plugin owners to determine which version of TGMPA has been loaded. Possibly useful for us as well for v3.0.


GaryJones:
Merge conflict.

jrfnl:
Rebased

PR #395, opened 2015-05-06
Feature/fix travis scrutinizer issues

jrfnl:
Fixes one actual bug, fixes the travis run issues and should also fix about half the Scrunitizer run issues.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -381,7 +381,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			/* After this point, the plugins should be registered and the configuration set. */
 
 			// Proceed only if we have plugins to handle.
-			if ( ! is_array( $this->plugins ) || empty( $this->plugins ) ) {
+			if ( empty( $this->plugins ) || ! is_array( $this->plugins ) ) {
 				return;
 			}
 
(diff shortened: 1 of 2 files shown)


class-tgm-plugin-activation.php
```diff
@@ -534,7 +534,9 @@ public function admin_init() {
 
 				global $tab, $body_id;
 				$body_id = 'plugin-information';
+				// @codingStandardsIgnoreStart
```
GaryJones: Why does this bit need to be ignored?

class-tgm-plugin-activation.php
```diff
@@ -2095,7 +2095,7 @@ protected function _gather_plugin_data() {
 				if ( ! empty( $upgrade_notice ) ) {
 					$table_data[ $i ]['upgrade_notice'] = $upgrade_notice;
 
-					add_action( "tgmpa_after_plugin_row_$slug", array( $this, 'wp_plugin_update_row' ), 10, 2 );
+					add_action( 'tgmpa_after_plugin_row_' . $slug, array( $this, 'wp_plugin_update_row' ), 10, 2 );
```
GaryJones: This bit needs reverting. Interpolation should be used for hooks, so that WP-Parser can correctly pick up the name for documenting the hooks. We can probably turn this check off for Scrutinizer if necessary.

https://core.trac.wordpress.org/ticket/26869#comment:29

class-tgm-plugin-activation.php
jrfnl: This is an alternative way to get round that one issue where we were considering excluding the rule or lowering the severity.
Now the rule is still being used, just not for that one line.

class-tgm-plugin-activation.php
jrfnl: Will do.
Though WP Parser should be fixed in that case - I believe my WP Hook Documentor can actually deal with both ;-)
Re: Scrutinizer - we can probably just set them to ignore in their interface for these lines without changing the config.

PR #397, opened 2015-05-08
PHPCS-related tweaks

GaryJones:
Originally branched to support the docs branch of WPCS, most of the issues have already been addressed in PRs already merged into dev.

What remains then is a readme update, and a few small tweaks.


.travis.yml
```diff
@@ -40,14 +40,8 @@ script:
...
     # @link https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards
     # @link https://github.com/squizlabs/PHP_CodeSniffer
-    # -p flag: Show progress of the run.
-    # -s flag: Show sniff codes in all reports.
-    # -v flag: Print verbose output.
-    # -n flag: Do not print warnings (shortcut for --warning-severity=0)
-    # --standard: Use WordPress (not VIP) as the standard.
-    # --extensions: Only sniff PHP files.
-    # --report: Show a summary report.
-    - $PHPCS_DIR/scripts/phpcs -p -s -v -n . --standard=./ruleset.xml --extensions=php --report=summary
```
jrfnl: The -v and -n flag are currently not included in the ruleset file.

class-tgm-plugin-activation.php
```diff
@@ -1519,7 +1519,9 @@ protected function get_plugins_api( $slug ) {
 
 				if ( is_wp_error( $response ) ) {
 					if ( true === WP_DEBUG ) {
+						// @codingStandardsIgnoreStart
 						wp_die( esc_html( $this->strings['oops'] ) . var_dump( $api ) ); // WPCS: xss ok.
+						// @codingStandardsIgnoreEnd
 					} else {
```
jrfnl: This will ignore that line always. As it only generates a warning and it's a valid warning which we may need to revisit, ignore warnings in travis and let this show up in manual/local runs.

GaryJones: Travis and local runs use the same config as provided by phpcs.xml. 

If we ignore warnings in Travis / local runs, we'll be ignoring all warnings, even ones we could easily fix, in code that isn't just for debugging. I'd rather just bypass this, and if we ever decide to review which code standards we're ignoring, this is an easy string to search for.

Alternatively, we just drop the `var_dump()` and wrapping condition altogether

.travis.yml
GaryJones: They don't need to be. `-n` is for ignoring warnings (which we shouldn't do), and `-v` is for verbose, and I've not found that to be any more useful than what the full report already provides.

Travis and local runs use the same config as provided by phpcs.xml.

class-tgm-plugin-activation.php
jrfnl: +1 on dropping of the `var_dump()`

Re - local runs - I run with a stricter/adjusted ruleset locally (as I presume you do), so that only applies for other contributors

PR #410, opened 2015-05-16
Update example code

grappler:
The old example code was missing some text domains


jrfnl:
Hi @grappler, thanks for keeping us sharp. I'm actually completely re-doing the gh-pages/website in preparation for the 2.5.0 release and this is one of the things which will be changed. I'll probably send in the PR for the re-do in a day or two. Would appreciate your feedback when I do.

grappler:
Ok, great. The reason I created this PR was that the old example was missing text domains and a number of theme authors use the example from the site.

Look forward to 2.5.0

jrfnl:
@grappler I've just send in the PR for the website #415. Your feedback would be appreciated.

PR #417, opened 2015-05-20
Fix URLs in custom Menus

### LINKED ISSUE #412: Begin installing plugins - 404
I am testing v2.5 and I am getting 404 when click on "Begin Installing Plugins".
I changed "parent_slug" value to custom value: "pixely_options" (I want to move "Install plugins" under my custom Menu) and now URL for installing plugins is wrong:

http://localhost/wordpress-latest/wp-admin/pixely_options?page=install-required-plugins&plugin_status=install

In previous versions there was "parent_url_slug" parameter that now seems not to be available and instead of it "parent_slug" is used.

I am doing something wrong or this is bug?

Vladimir

jrfnl:
Fixes #412

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1630,11 +1630,15 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			static $url;
 
 			if ( ! isset( $url ) ) {
+				$parent = $this->parent_slug;
+				if ( false === strpos( $parent, '.php' ) ) {
+					$parent = 'admin.php';
+				}
 				$url = add_query_arg(
 					array(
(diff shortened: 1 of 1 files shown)


class-tgm-plugin-activation.php
```diff
@@ -1630,11 +1630,15 @@ public function get_tgmpa_url() {
 			static $url;
 
 			if ( ! isset( $url ) ) {
+				$parent = $this->parent_slug;
+				if ( strpos( $parent, '.php' ) === false ) {
```
GaryJones: Yoda condition please.

jrfnl: You'll just have to keep telling me that again and again... I just can't get used to it. Anyways, fixed & squashed.

PR #427, opened 2015-06-03
Feature/lean load better errors

jrfnl:
Two more lean loading fixes and better error messages in case the user chooses a bulk action, but doesn't select any plugins.


jrfnl:
The travis fail is due to new WPCS release - see PR #426

PR #429, opened 2015-06-04
Fix notice appearing in WP 4.3-alpha

GaryJones:
Due to https://github.com/WordPress/WordPress/commit/4c0c7fd7ba401869682008ee4dc87a9c26d8a262, WP 4.3 now expects `$this->_column_headers` property in the list table object to have 4 entries. This commit fixes the offset notice that otherwise appears.

In WP 4.2, the `list()` call inside `single_row_columns()` simply drops the extra values, so this is backwards compatible.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -2527,6 +2527,18 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
 		}
 
 		/**
+		 * Get name of default primary column
+		 *
+		 * @since 2.5.0
+		 * @access protected
+		 *
+		 * @return string
+		 */
+		protected function get_default_primary_column_name() {
+		    return 'plugin';
+		}
+
+		/**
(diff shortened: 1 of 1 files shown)


GaryJones:
@jrfnl Any clue why we might be manually populating the `$_column_headers` property instead of letting WP handle that?

jrfnl:
> Any clue why we might be manually populating the $_column_headers property instead of letting WP handle that?

At first glance - because the normal way of populating it involves screen object related functions (`get_column_headers()`, `get_hidden_columns()`) which would then need to be filtered or set.

GaryJones:
This may be one we can address in 3.0.0 then.

jrfnl:
We do need to have something left to do *after* 3.0 as well though ;-)

jrfnl:
Looks like there is some fall-out from this: `Fatal error: Call to undefined method TGMPA_List_Table::get_primary_column_name() ` in WP 3.9.

jrfnl:
Should be fixed by PR #432

PR #433, opened 2015-06-14
Fix 'update' link always showing in notice…

jrfnl:
... even when no updates available.

Also: redirect to 'all' view when selected view does not have any plugins to display.

Re bug reported over skype:
> Gary Jones: Only thing is that Begin updating plugins link in the admin notice - when there are none to update. Clicking that gives you an empty table, and a notice that there are "No plugins to install, update, or activate", yet the filter links above says "All (4) | To Install (2) | To Activate (2)". A bit confusing.
Or, if there's nothing to update, we redirect to showing All instead.
screenshot

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1038,9 +1038,10 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				}
 
 				if ( ! $this->is_plugin_installed( $slug ) ) {
-					$install_link_count++;
 
 					if ( current_user_can( 'install_plugins' ) ) {
+						$install_link_count++;
+
 						if ( true === $plugin['required'] ) {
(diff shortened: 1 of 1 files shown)


GaryJones:
Can't merge.

The merge I just did from you had the whole file replaced too - some tabs vs spaces or line-endings issue?

GaryJones:
Fine to merge actual intended changes once sorted.

jrfnl:
> some tabs vs spaces or line-endings issue?

Travis borked on some obscure line ending issue. I commit Unix-style, but it still did not like the file. Anyways, rebased.

GaryJones:
My point was that https://github.com/TGMPA/TGM-Plugin-Activation/pull/432/files has a stupid number of changes (so that was the bad commit), and this one was fine as it was reverting whatever the tabs / spaces / line-endings issue was in #432.

PR #435, opened 2015-06-15
Code tidy

GaryJones:
* Remove redundant clear lines when changing to a different level of indentation.
* Add a clear line before `return` when it follows other code at the same level of indentation.
* Fix or add some missing `@return` docs.
* Change plugin headers to use real names.
* Change version to `-RC` since we're not yet ready to tag.
* Remove space between `@param` and `@return` per WP coding standards.
* Remove incorrect alignment of some `@return` descriptions.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -8,7 +8,7 @@
  * or theme author for support.
  *
  * @package   TGM-Plugin-Activation
- * @version   2.5.0
+ * @version   2.5.0-RC
  * @link      http://tgmpluginactivation.com/
  * @author    Thomas Griffin, Gary Jones, Juliette Reinders Folmer
  * @copyright Copyright (c) 2011, Thomas Griffin
@@ -18,8 +18,8 @@
  * Plugin Name: TGM Plugin Activation
  * Plugin URI:
  * Description: Plugin installation and activation for WordPress themes.
- * Version:     2.5.0
- * Author:      griffinjt, garyj, jrf
+ * Version:     2.5.0-RC
+ * Author:      Thomas Griffin, Gary Jones, Juliette Reinders Folmer
  * Author URI:  http://tgmpluginactivation.com/
  * Text Domain: tgmpa
  * Domain Path: /languages/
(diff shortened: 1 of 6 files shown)


corvannoorloos:
Should the `.travis.yml` file also be re-indented with 2 instead of 4/mixed spaces?

class-tgm-plugin-activation.php
```diff
@@ -18,8 +18,8 @@
  * Plugin Name: TGM Plugin Activation
  * Plugin URI:
  * Description: Plugin installation and activation for WordPress themes.
- * Version:     2.5.0
- * Author:      griffinjt, garyj, jrf
+ * Version:     2.5.0-RC
+ * Author:      Thomas Griffin, Gary Jones, Juliette Reinders Folmer
```
jrfnl: This is WP Plugin repo compliant header and should contain the WP.org nicks for the authors, so please revert.

class-tgm-plugin-activation.php
```diff
@@ -8,7 +8,7 @@
  * or theme author for support.
  *
  * @package   TGM-Plugin-Activation
- * @version   2.5.0
+ * @version   2.5.0-RC
```
jrfnl: +1

class-tgm-plugin-activation.php
```diff
@@ -457,11 +450,11 @@ public function add_plugin_action_link_filters() {
 		 * @since 2.5.0
 		 *
 		 * @param array $actions Action links.
-		 *
```
jrfnl: I seem to remember that we both said we preferred to have the line between @param's and @return lines, so why not do so ?
This is not WP Core after all (yet) and the empty line between @param and @return also complies with all the examples included in the upcoming PSR-5.

class-tgm-plugin-activation.php
GaryJones: Incorrect - the author names in the plugin headers are used on the Plugin list page - human readable.

WP.org usernames would only be needed for a readme.txt so that the repo links them correctly.

class-tgm-plugin-activation.php
GaryJones: I've since chatted to @DrewAPicture and he's confirmed that the WP standard is to group `@param` and `@return` together (as oer the WP CS handbook), so that trumps our preference and PSR-5.

GaryJones:
@corvannoorloos Yes, you're probably right.

PR #441, opened 2015-07-02
Simplify list logic with wp_sprintf()

GaryJones:
Removes an ugly translation string too.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1113,8 +1113,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 
 					$count          = count( $plugin_group );
 					$linked_plugins = array_map( array( 'TGM_Utils', 'wrap_in_em' ), $linked_plugins );
-					$last_plugin    = array_pop( $linked_plugins ); // Pop off last name to prep for readability.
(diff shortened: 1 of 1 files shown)


jrfnl:
I'm going to have to study this `wp_sprintf()` function to figure out how the heck this works, but it *does* seem to work.

Two notes:
* Using this function is 'heavier' than the previous implementation
* The result might be mixed language which I don't think is a good thing usability-wise. If there is a WP translation for a language, but no TGMPA translation, you'll end up with a list saying "English TGMPA phrase: Plugin A, Plugin B _translated and_ Plugin C".

GaryJones:
Agreed on both points, however:

* It lets WP handle the logic (yours was sound, as far as I know), and allows it to be filtered accordingly via the filters in `wp_sprintf()` and `wp_sprintf_l()`. If we're looking to put this in core, then using existing functions, even heavier ones, would surely be more preferable.
* Even if it's just one word translated, that's better than none.

jrfnl:
> If we're looking to put this in core, then using existing functions, even heavier ones, would surely be more preferable.

True, but that would be after the rewrite.

> Even if it's just one word translated, that's better than none.

I don't agree, having half a sentence in one language and half in another, makes for the need for brain switches (between languages) which will make the sentence harder to read. Also gives problems with sentences where one half is LTR (english) and the other bit is RTL (arab).

PR #446, opened 2015-07-13
Feature/rename class

jrfnl:
Rename the TGM_Utils class to TGMPA_Utils as it conflicts with the Soliloquy plugin using the same classname.

Upped all version numbers, i.e. PR is prepped for release straight away.


jrfnl:
See also: https://github.com/inboundnow/landing-pages/issues/178

PR #448, opened 2015-07-15
Feature/236 compatibility

jrfnl:
For immediate release.

Fixes compatibility issues when several TGMPA libaries are included in a single WP install (both themes as well as one or more plugins using TGMPA).

Addresses the following issues:
* `Fatal error: Call to protected TGM_Plugin_Activation::__construct()` error when v2.3.6 or lower would be loaded after v2.5.0
* Incorrect setting of `menu` property when v2.3.6 or lower would be loaded after v2.5.0
* Incorrect using of pre-2.5 bulk installer classes when v 2.5+ is the loaded version and v2.4.1 or lower would be loaded after v2.5.0.
* Strict standards errors when v2.4.1 or lower would be loaded after v2.5.0.
* Errors for bulk installer classes when v 2.5+ would be loaded after an earlier version.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -149,7 +149,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 *
 		 * @var string
 		 */
-		public $menu = 'tgmpa-install-plugins';
+		protected $menu = 'tgmpa-install-plugins';
 
 		/**
 		 * Parent menu file slug.
@@ -256,11 +256,15 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 * Adds a reference of this object to $instance, populates default strings,
 		 * does the tgmpa_init action hook, and hooks in the interactions to init.
 		 *
+		 * @internal This method should be `protected`, but as too many TGMPA implementations
+		 * haven't upgraded beyond v2.3.6 yet, this gives backward compatibility issues.
+		 * Reverted back to public for the time being.
+		 *
 		 * @since 1.0.0
 		 *
 		 * @see TGM_Plugin_Activation::init()
 		 */
-		protected function __construct() {
+		public function __construct() {
 			// Set the current WordPress version.
 			$this->wp_version = $GLOBALS['wp_version'];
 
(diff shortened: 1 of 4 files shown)


class-tgm-plugin-activation.php
```diff
@@ -272,6 +276,34 @@ protected function __construct() {
...
+		 *
+		 * @see https://github.com/TGMPA/TGM-Plugin-Activation/blob/2.3.6/tgm-plugin-activation/class-tgm-plugin-activation.php#L1593
+		 *
+		 * @param string $name  Name of an inaccessible property.
+		 * @param mixed  $value Value to assign to the property.
+		 * @return void  Silently fail to set the property when this is tried from outside of this class context.
+		 *               (Inside this class context, the __set() method if not used as there is direct access.)
+		 */
+		public function __set( $name, $value ) {
+			return;
```
GaryJones: Do we want to limit this to only the `$menu` property? Setting it for all properties may potentially cause issues later on - limiting the effect of this hackedihack may be better.

class-tgm-plugin-activation.php
```diff
@@ -272,6 +276,34 @@ protected function __construct() {
...
+		}
+
+		/**
+		 * Magic method to get the value of a protected property outside of this class context.
+		 *
+		 * @param string $name Name of an inaccessible property.
+		 * @return mixed The property value.
+		 */
+		public function __get( $name ) {
```
GaryJones: What's the purpose of this method? Isn't it just returning the default value anyway? I don't think you need to specify `__get()` just because there's a `__set()` defined.

class-tgm-plugin-activation.php
jrfnl: I considered that, but figured the risk is low as development will probably move towards 3.0 now and 3.0 will look quite differently internally.

Anyway, the magic method `__set()` (and `__get()` for that matter) is only called for protected and private properties and only when someone tries to set those from outside the class. Methods within the class have direct access. Same goes for public properties from outside the class (direct access).

In other words, these methods will currently only be called for `$sort_order`, `$has_forced_activation`, `$has_forced_deactivation` and now `$menu`.
The first three should not be changed outside the class anyway (though could be from a child class, aka `protected` rather than `private`).

PR #459, opened 2015-08-19
Action links would not always be filtered.

jrfnl:
Was hooked in only when not all plugins were active/installed/up-to-date, while the filters to prevent deactivation of force-activated plugins should also be added when all required/recommended plugins are already there.


shivapoudel:
+1

PR #469, opened 2015-08-28
Use correct directory path

emiluzelac:
`get_template_directory()` should be used in both cases. 

- get_template_directory(): Returns the absolute template directory path.
- get_stylesheet_directory(): Returns the absolute stylesheet directory path.

Maybe you could recommend https://wordpress.org/plugins/example/ instead of one that bundles.

### DIFF
--- a/example.php
+++ b/example.php
@@ -20,7 +20,7 @@
 /**
  * Include the TGM_Plugin_Activation class.
  */
-require_once dirname( __FILE__ ) . '/class-tgm-plugin-activation.php';
+require_once get_template_directory() . '/class-tgm-plugin-activation.php';
 
 add_action( 'tgmpa_register', 'my_theme_register_required_plugins' );
 /**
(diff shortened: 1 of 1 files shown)


GaryJones:
There's several things wrong here:

* You're assuming that TGMPA is being used within a theme. For plugins that use TGMPA, `get_template_directory() . '/class-tgm-plugin-activation.php'` won't be valid.
* You're assuming that `class-tgm-plugin-activation.php` is within the top-level of a theme. It's far more likely, and therefore a better default, to find a tgmpa setup file within the same directory as the class, regardless of how many directories deep that may be. 
* Even if TGMPA was only for themes, child themes may want to include it when parent themes don't have so `get_stylesheet_directory()` would be more accurate anyway, which is why we use it in your second change.

emiluzelac:
My bad, I wasn't thinking about the plugins at all :(

As far as `dirname( __FILE__ )` maybe you could add a note somewhere that this would need to be changed accordingly if included with themes that is. This is a required change for .org and ThemeForest as well.

GaryJones:
How about you do a PR which has commented-out versions suitable for plugin and theme, replacing what's currently there?

e.g.
```php
// Uncomment ONE of the following typical paths and amend as necessary.
// Theme:
// require get_stylesheet_path() . '/path/to/class-tgm-plugin-activation.php';
// Plugin:
// require {whatever the plugins_dir_path() thing is that I'm too lazy to look up right now  :-) };
```

emiluzelac:
I'll try, time isn't on my side lately :)

Issue #475, opened 2015-09-03
add_theme_page() Theme Check error


galengidman:
When running Theme Check on a theme using version 2.5.2, I get the following error...

```
REQUIRED: tgm-plugin-activation/class-tgm-plugin-activation.php. Themes should use add_theme_page() for adding admin pages.

`Line 646: $this->page_hook = call_user_func( 'add_{$type}_page', $args['parent_slug'], $args['page_title'], $args['menu_t ...
```

Since the wordpress.org uploader checks the theme before allowing an upload, this makes it so that I cannot even upload my theme. Any fixes for this coming?

Otto42:
This is a fork that passes theme-check and removes the undesirable ability to add arbitrary admin menus.

https://github.com/Otto42/TGM-Plugin-Activation

kisonic:
I have this error too:
REQUIRED: plugins/tgm-plugin-activation/class-tgm-plugin-activation.php. Themes should use add_theme_page() for adding admin pages.
Line 624: ```$this->page_hook = call_user_func( 'add_submenu_page', $args['parent_slug'], $args['page_title'], $args['menu_t```

jrfnl:
Closing as the [Custom TGMPA Generator](http://tgmpluginactivation.com/download/), which is now available on the website, fixes this.

PR #476, opened 2015-09-04
updated add_submenu_page call to work around Theme Check

galengidman:
Did some digging and found a solution to #475. I guess Theme Check updated at some point which made the 9896b44 fix no longer work.


shivapoudel:
@GaryJones if set for theme in config then let's by default point submenu to appearance menu, but if it was for plugin then let the user decide. If there exist some logic to check tgmpa is in plugin or in theme directory then it will be smart way to handle those config too.

jrfnl:
Theme Check would just be adjusted to detect this. Closing - downward spiral. See discussion in #460.

galengidman:
Ah, didn't see that. Thanks for pointing me to it.

jrfnl:
FYI: The  [Custom TGMPA Generator](http://tgmpluginactivation.com/download/), which is now available on the website, fixes this.

PR #478, opened 2015-09-04
Fixed: Wrong class for notification box

Ninos:


jrfnl:
The `update-nag` class was introduced in 3.9, you are now disabling it effectively. Any particular reason ?

Ninos:
huh really? Because of following styling:
tgmpa - notice box

WP v4.3

jrfnl:
I don't get it - what do you mean ? Looks ok to me, though maybe we should disable the dismiss button or remove our own dismiss text if on 4.3 or up (now double with different logic behind it).

Ninos:
The box does not use the full line / width.

Ninos:
Ok I don't konw in which version they changed the class name back to "updated", but in v4.3 wordpress is definitely using `updated` instead of ´update-nag´.

I've commited some changes.

jrfnl:
Hi @Ninos I finally had time to properly test this and ended up with a slightly different implementation #509 for consistency. Thank you for bringing this to our attention and I look forward to your next pull request.

Issue #236, opened 2014-10-30
Fatal error when using composer


entr:
I'm getting `Fatal error: Call to undefined function __() in ...\vendor\tgm\plugin-activation\class-tgm-plugin-activation.php on line 161` when using like composer package. What do i miss?

chetzof:
I got this error when included autoloader from wp-config.php
However, when i moved the inclusion into theme's functions.php  - error has been fixed.

But obviously would be great to get this working out of the box.

entr:
This makes sense. Not sure if theme's functions.php is the best place to run autoloaded tough. I would hook to `wp_loaded` or `init` and autoload there.

GaryJones:
I guess at the point of wp-config.php, the `__()` and some other WP functions have not yet been defined.

Pinging @Rarst for advice.

Rarst:
Top of `wp-config.php` is common place at the moment, which I recommend myself. Autoload should not do be made to do anything that would break if loaded before WP core in this or other way.

GaryJones:
So the autoloading is pulling in `class-tgm-plugin-activation.php`, but that's just defining classes and functions. What's being called to trigger the error?

marcinant:
:+1:  bump. Is there any workaround?

jrfnl:
@marcinant We need more information to answer this. Please answer the questions asked above and preferably provide us with a [backtrace of the PHP error](https://gist.github.com/jrfnl/5925642).

GaryJones:
Closing due to lack of feedback.

renegeuze:
The problem remains and I'm happy to give some explanation.

The class-tgm-plugin-activation really did just define classes and functions like  `did_action` on #L1915
That works great as long as you include the autoloader somewhere after the big initialisation of Wordpress functions but does not work when including it earlier.

That concludes the part of what you already know and maybe even intended.

When you setup Wordpress with for example, [Bedrock](https://roots.io/bedrock/), and put tgmpa in the root composer file it will be loaded too early and you'll have no control over it.

Unfortunately I have no ready to go solution that doesn't break backwards compatibility except for skipping tgmpa and use wppackagist for plugins without tgmpa. But you'd lose the nice features tgmpa has.

Example error trace:

	( ! ) Fatal error: Call to undefined function did_action() in [...]/vendor/tgmpa/tgm-plugin-activation/class-tgm-plugin-activation.php on line 1910
	Call Stack
[...]

PR #488, opened 2015-10-13
Php error when auto-loading class prior to Wp Themes or Plugins initialization

tlartaud:
prevent error **Call to undefined function did_action() in class-tgm-plugin-activation.php**

https://www.google.fr/search?q=Call%20to%20undefined%20function%20did_action()%20in%20class-tgm-plugin-activation.php&rct=j


jrfnl:
@rarst Would you mind reviewing this for me ? You're so much better at Composer ;-)

tlartaud:
oh sorry, I didn't meant to be the boss. I apologize if i am misunderstanding something, but,

I have a project built on a WP stack which is including a composer.json.
According to what I understand, the `vendor/composer/autoload_files.php` can be loaded inside a wp-plugin, inside or wp-theme, or prior to wp initialization where WP is just a composer dependency.

Could you maybe explain me why I'm facing this problem ?

jrfnl:
@tlartaud Hi Thomas, no need to apologize. I understand the issue the you are facing and appreciate very much that you are sending in a PR for this.

As I personally don't use Composer, I figured I'm not the best person to review this PR, aka my request for help to someone who I know is very well versed in Composer in combination with WP.

What strikes me as odd about the error is that autoload would load the files before WP. TGMPA should always be a dependency of a plugin or theme not of a WP install, as it won't do anything without a function in the plugin or theme providing it with the dependencies (and WP core does not contain such a function).
So I'm wondering if this Composer setup is appropriate.

Autoloading the file when it's part of a vendor directory for a plugin or theme seems good to me, autoloading it as part of WP not so much. But then again, as I said, I don't use Composer, so I'd like the opinion of someone who knows more about this sort of setups.

Rarst:
The file in question is mixing definitions with runtime code. It _will_ fall apart regardless of _how_ you call it, if it's done too early.

So this is architectural issue, not just autoload one. I am not confident what would make sense, not having used the library in practice. As far as I am concerned any Composer-enabled WP project should work as part of site stack, otherwise bulk of value Composer brings goes out of the window.

In general I would recommend to untangle definitions from runtime code, PSR-1 et al.

jrfnl:
@Rarst Thanks for your input!

renegeuze:
Accepting this code will break the plugin for people who use composer autoloading from inside functions.php or any other late loading file.
Of course nobody should have done that but it was an available option for a long time so I'm sure somebody did.

Better solution is to change the entire structure(as @Rarst mentions) and bump up the version.
For a backwards compatible fix, wrap parts inside function_exist. Also see: https://github.com/TGMPA/TGM-Plugin-Activation/issues/236#issuecomment-149845637

GaryJones:
I've hit the same issue, when trying to run theme-level PHPCS or PHPUnit, on a theme that pulls in TGMPA via Composer. I load theme files by requiring `vendor/autoload.php`, so it naturally loads up the TGMPA class file too.

One neater solution, seems to be to add the following to the top of TGMPA:

```php
if ( ! defined( 'ABSPATH' ) ) {
	return;
}
```

@Rarst @jrfnl Any drawbacks you can see with that?

jrfnl:
@GaryJones There's a PR for that were I asked the same question: #594

GaryJones:
#594 is merged.

PR #495, opened 2015-11-29
Fix the nag style on wide screens

primozcigler:
Before: http://www.awesomescreenshot.com/image/792720/2269bfceda02f35bed2c0fe2960e5397
Now (fixed): http://www.awesomescreenshot.com/image/792721/ba77c7991e14a179858e601faf2f3673

The problem was `update-nag` class which has the `display: inline-block;` applied.


primozcigler:
I see #478 is related, but the actual implementation there appears to be wrong to me.

jrfnl:
Hi @primozcigler I finally had time to properly test this and ended up with a slightly different implementation #509 for consistency. Thank you for bringing this to our attention and I look forward to your next pull request.

primozcigler:
:+1:

PR #502, opened 2016-01-04
Update travis script.

jrfnl:
* Include PHP 7 in the matrix
* Fair use of travis/faster finish: sniff results won't change across PHP versions, so only run once.


jrfnl:
Will check what's wrong & fix once I can see what's up. Travis CDN seems to be down, so can't see any build results. And it just has to be said: I *detest* sites which are 100% ajax... 0% accessability....

PR #503, opened 2016-01-04
Updated the example file with examples for different include situations.

jrfnl:
Fixes #469.

### DIFF
--- a/example.php
+++ b/example.php
@@ -19,10 +19,22 @@
 
 /**
  * Include the TGM_Plugin_Activation class.
+ *
+ * Depending on your implementation, you may want to change the include call:
+ *
+ * Parent Theme:
+ * require_once get_template_directory() . '/path/to/class-tgm-plugin-activation.php';
+ *
+ * Child Theme:
+ * require_once get_stylesheet_directory() . '/path/to/class-tgm-plugin-activation.php';
+ *
+ * Plugin:
+ * require_once dirname( __FILE__ ) . '/path/to/class-tgm-plugin-activation.php';
  */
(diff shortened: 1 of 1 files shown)


example.php
```diff
@@ -19,10 +19,22 @@
 
 /**
  * Include the TGM_Plugin_Activation class.
+ *
+ * Depending on your implementation, you may want to change the include call:
+ *
+ * Parent Theme:
+ * require_once get_template_directory() . '/path/to/class-tgm-plugin-activation.php';
```
GaryJones: No need for this - parent themes can call `get_stylesheet_directory()` too, since template dir = stylesheet dir in parent themes.

jrfnl: True, however, not applicable in this case.

If no child theme is being loaded, the `get_stylesheet_directory()` would be fine as WP will detect that there is no child theme and return the parent theme directory.

If a parent theme uses TGMPA and a child theme of that parent is the loaded theme, things would go very wrong with the include call. This is not a url/path which is being passed to a `get_template_part()` call, after all, but a hard file include.

To demonstrate, take this example set up:
```
/wp-content/themes/
	- /parent-theme
		- /vendor/tgmpa/tgm-plugin-activation.php
	- /child-theme
		- does not include tgmpa
```

The `get_template_directory()` function would return `/path/to/wp-content/themes/parent-theme`.
The `get_stylesheet_directory()` function would return `/path/to/wp-content/themes/child-theme`.

If the parent theme would include the file using `get_template_directory()`, this would result in a fatal `file not found` error.

Issue #489, opened 2015-10-13
Don't prompt for plugin updates


ollietreend:
Hi,

It would be great to have an configuration option which disables the *'There is an update available'* prompt for plugins. It would be even better if this could be selectively disabled, based on whether the user has permission to install plugin updates.

One problem I'm currently faced with is that I have non-administrator users using the WordPress admin interface. If there are updates available for required plugins, they see a prompt which says *'Sorry, but you do not have the correct permissions to update the [...] plugin.'*

This message appears at the top of every page for my non-administrator users, until an administrator logs in and performs the necessary updates. Not only is the wording of this message confusing for users – because it suggests that the user has tried to perform an action (update plugins) which they didn't – but it also adds an annoying nag which appears at the top of every page.

<img width="375" alt="Screenshot of update nag" src="https://cloud.githubusercontent.com/assets/7735945/10454671/5b40150c-71ae-11e5-953a-0d96e698a011.png">

In this screenshot I am logged in as an Editor. I have simply loaded the 'All Pages' list, and haven't attempted to perform any plugin updates. But I am greeted with a warning message to tell me that I don't have permission. This is not a good user experience.

I've looked into the TGM class to resolve this, but I can't see any convenient actions or filters for me to hook in to. Nor can I find a configuration option to disable update nags.

Your thoughts and ideas for implementation would be great. I'm thinking either a configuration option, or the ability to hook into a filter (or even both) would provide suitable flexibility. Beyond that, perhaps there is no need to show the 'Sorry' message at all when a user doesn't have permission to perform updates – after all, they didn't try to install anything, so why should they see the message?

jrfnl:
Related #414

cjgrasso:
This was posted a year ago, is there any update on this?  Every user role is getting the plugin installation and activation nags even though these should only be showed to administrators.

thanks

jrfnl:
@cjgrasso A lot was done - see PR #507.

cjgrasso:
@jrfnl thank you for the help.

paulcanning:
@jrfnl how do i actually go about disabling these notices for non-super-admins?

jrfnl:
@paulcanning You hook into the `tgmpa_show_admin_notice_capability` filter and return a different capability.

[...]

Issue #414, opened 2015-05-18
No permissions admin notice


GaryJones:
On logging in to another site where I'm not an administrator, I saw:

![screenshot 2015-05-18 22 18 30](https://cloud.githubusercontent.com/assets/88371/7690502/f39e4426-fdab-11e4-8fda-ea28fc7365a1.png)

While it's obviously good that a normal user can't install plugins, I don't think this message serves any purpose in being shown at all. As a user, I should not need to know what plugins a site should be running. Let's hide the message altogether and remove the associated strings that are currently being shown, and return early from any admin notice functions.

jrfnl:
That completely depends on the kind of site. For a large multi-user site, I can see your point.
But for the more common - dev = admin, and three or four users which are just below admin, the message is useful. The dev might only be involved at the customers request and not keep up with which plugins need updating and such. Users need to know to contact the dev to get things working properly.

What about if we add a capability-check for which users can see it ? Something like author minimum ?

GaryJones:
> Users need to know to contact the dev to get things working properly.

That's common sense already. 

I guess the problem is that the details of which plugins are shown, when the user can't do anything about _any_ plugins. How about author minimum, but also make the text generic like "This site requires one or more plugins that are not installed. Contact the ..."?

jrfnl:
May be we should also consider to *only* show the notice if *required* plugins need to be installed or updated. And not show anything at all to non-admins if the notice would only deal with recommended plugins.

jrfnl:
And now that updating is in - if the user has dismissed the message before, it would not show for new updates anymore. That kind of feels wrong as well.

Basically I think we may need to rename issue #353 to "Refactor/rethink complete notices logic" and collect all the different bits there.

And we need to think through whether that should something we still want to do for 2.5. It *is* already on the roadmap for 3.0

MickeyKay:
I'm running into the same issue. My thought is that it would make sense to set the minimum user role or capability for which the notice should be displayed. That'd be a pretty simple fix.

GaryJones:
Definitely not a 2.5.0 thing. Either add a capability check in 2.6.0, or rewrite completely for 3.0.0

PR #507, opened 2016-01-05
Variable admin notices based on user level.

### LINKED ISSUE #190: Filter for disabling the notice box?
Is it possible to add a filter for disabling the notice box? We're working on a multisite installation with a parent theme that uses TGM and we don't want it to show up on every admins profiles.

We have hooked into get_user_metadata to override the default setting value globally, but it would be better if there was a native filter available.

Here's the code we are using now:

```php
add_filter('get_user_metadata', function($val, $object_id, $meta_key, $single)
{
    if($meta_key === 'tgmpa_dismissed_notice')
        return true;
    else
        return null;
 
}, 10, 4);
```

More info:
http://snippets.khromov.se/disable-tgm-plugin-activation-nag-screen/

### LINKED ISSUE #479: Dismiss update notice
I have required a plugin from wordpress repo and I have set 'dismissable' to false.But when is a new update I don't want to bother the user on every single page to update the plugin that I have required.How can I do this?

### LINKED ISSUE #492: Have more than 1 TGMPA backend notification
Hi,

I'm wondering if there would be a way to include 2 separate notifications in the backend. 1 for the User and One For Developer as i'm creating a Theme Framework.

Thanks Jake.

jrfnl:
By default the following logic will be used:
User < `'publish_posts'` (=Author): no admin notices
User < install/update/activate: _"contact administrator notice"_ without disclosing information about the plugins involved and only if there are *required* plugins which require action.
User = install/update/activate: "normal" notices

The minimum user level for which admin notices are shown at all can be adjusted by using the newly introduced `tgmpa_show_admin_notice_capability` filter. The default capability is set to `'publish_posts'`.

Example:
```php
add_filter( 'tgmpa_show_admin_notice_capability', create_function( '$cap', 'return \'edit_pages\';' ) );
```

To only show the admin notices to network admins on multisite, set it to a super admin capability like `'manage_network_plugins'`.

Note: the `notices()` function is ugly and in desperate need of refactoring, that is not handled in this PR (which only makes it worse).

Fixes #190, #414
Supersedes: https://github.com/INN/Largo/pull/740/files
Partially fixes #479, #489 - notice will now only show for required updates for non-admin users with level author or editor.
Possibly fixes #492, though more information is needed on the actual case.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -333,11 +333,6 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 					'This theme recommends the following plugins: %1$s.',
 					'tgmpa'
 				),
-				'notice_cannot_install'           => _n_noop(
-					'Sorry, but you do not have the correct permissions to install the %1$s plugin.',
-					'Sorry, but you do not have the correct permissions to install the %1$s plugins.',
-					'tgmpa'
-				),
 				'notice_ask_to_update'            => _n_noop(
 					'The following plugin needs to be updated to its latest version to ensure maximum compatibility with this theme: %1$s.',
 					'The following plugins need to be updated to their latest version to ensure maximum compatibility with this theme: %1$s.',
@@ -348,11 +343,6 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 					'There are updates available for the following plugins: %1$s.',
 					'tgmpa'
 				),
-				'notice_cannot_update'            => _n_noop(
-					'Sorry, but you do not have the correct permissions to update the %1$s plugin.',
-					'Sorry, but you do not have the correct permissions to update the %1$s plugins.',
-					'tgmpa'
-				),
 				'notice_can_activate_required'    => _n_noop(
 					'The following required plugin is currently inactive: %1$s.',
 					'The following required plugins are currently inactive: %1$s.',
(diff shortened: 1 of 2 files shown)


GaryJones:
The PR now has conflicts.

jrfnl:
Rebased.

class-tgm-plugin-activation.php
```diff
@@ -1179,6 +1127,65 @@ public function notices() {
...
+					translate_nooped_plural( $this->strings['activate_link'], $activate_link_count, 'tgmpa' ),
+					esc_url( $this->get_tgmpa_status_url( 'activate' ) )
+				);
+			}
+
+			$action_links = apply_filters( 'tgmpa_notice_action_links', $action_links );
+
+			$action_links = array_filter( (array) $action_links ); // Remove any empty array items.
+
+			if ( ! empty( $action_links ) && is_array( $action_links ) ) {
```
GaryJones: The `is_array()` check here is redundant - the `$action_links` have been cast to an array on the previous line.

class-tgm-plugin-activation.php
```diff
@@ -1179,6 +1127,65 @@ public function notices() {
 		}
 
 		/**
+		 * Generate the user action links for the admin notice.
+		 *
+		 * @since 2.x.x
+		 *
+		 * @param int $install_link_count  Number of plugins to install.
```
GaryJones: Could these arg names be improved? If it's the number of plugins to install, why is it making the presumption that it's going to be used in a link? Name the variable for what it is, i.e. `$installable_count` or `$installable_plugin_count`.

jrfnl:
@GaryJones re: both your remarks - for this bit, all I really did was abstract a block of code out to a new method as what with the other extra if's and indents, the `notices()` function became even more unwieldy than it already was. I didn't actually change the code within the block (other than adding a return).
So variable names, conditions etc are all as they were.

PR #512, opened 2016-01-07
Prevent TGMPA notice showing on WP core update pages.

jrfnl:
Prevents core/plugin/theme updates screens looking like this:

screenshot

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1001,7 +1001,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 */
 		public function notices() {
 			// Remove nag on the install page / Return early if the nag message has been dismissed or user < author.
-			if ( $this->is_tgmpa_page() || get_user_meta( get_current_user_id(), 'tgmpa_dismissed_notice_' . $this->id, true ) || ! current_user_can( apply_filters( 'tgmpa_show_admin_notice_capability', 'publish_posts' ) ) ) {
+			if ( ( $this->is_tgmpa_page() || $this->is_core_update_page() ) || get_user_meta( get_current_user_id(), 'tgmpa_dismissed_notice_' . $this->id, true ) || ! current_user_can( apply_filters( 'tgmpa_show_admin_notice_capability', 'publish_posts' ) ) ) {
 				return;
 			}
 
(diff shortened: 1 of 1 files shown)


class-tgm-plugin-activation.php
```diff
@@ -1601,6 +1601,33 @@ protected function is_tgmpa_page() {
...
 		/**
+		 * Determine if we're on a WP Core installation/upgrade page.
+		 *
+		 * @since 2.x.x
+		 *
+		 * @return boolean True when on a WP Core installation/upgrade page, false otherwise.
+		 */
+		protected function is_core_update_page() {
+			// Current screen is not always available, most notably on the customizer screen.
+			if ( function_exists( 'get_current_screen' ) ) {
```
GaryJones: Do an inverse check, and return false early.

class-tgm-plugin-activation.php
```diff
@@ -1601,6 +1601,33 @@ protected function is_tgmpa_page() {
...
+
+				$screen = get_current_screen();
+
+				if ( 'update-core' === $screen->base ) {
+					// Core update screen.
+					return true;
+				} elseif ( 'plugins' === $screen->base && isset( $POST['action'] ) ) {
```
GaryJones: Should the value of `action` be checked?

jrfnl: No, just the fact that it's set is indication enough that one of the WP bulk update screens will be displayed. `$_POST['action']` is only set when a bulk action is requested. (though I should correct the typo ;-) => done)

class-tgm-plugin-activation.php
jrfnl: Done

jrfnl:
I got push-back from WPCS about not checking the nonce for the `$_POST` - `WordPress.CSRF.NonceVerification.NoNonceVerification`. I don't think this opens up any CSRF issues in this case, so have chosen to whitelist, because:
a) it's not 'our' nonce to check.
b) we're not actually using the value
c) if for whatever nefarious reason the `action` variable is set & not empty, it would only cause the notice not to display, rather lessening any security exposure, rather than heightening it.

I did change the `isset()` to a `! empty()` as an empty `action` should not be a reason not to display the notice.
Let me know if you disagree.

GaryJones:
It's not our nonce, but if there is a nonce, we should still check it, since we're still wanting to confirm that we're on the plugins / update page, and that an action was posted *from within that page*. My making an external POST to example.com/wp-admin/plugins.php should not trigger TGMPA into doing anything.

We may not be using the value, but we are using the fact the key exists, to make an assumption.

Whitelisting is the wrong decision here IMO.

jrfnl:
> My making an external POST to example.com/wp-admin/plugins.php should not trigger TGMPA into doing anything.

That's the whole point why I *did* whitelist. In that case, we don't want to expose TGMPA and therefore shouldn't show the notice.

The alternative would be to check the nonce + rest and if ok, not show the notice and if not ok, also not show the notice, so the end result would be exactly the same, just with more code.

GaryJones:
Whitelisting doesn't affect exposure. Using the nonce does. The nonce is the right way to go here.

jrfnl:
@GaryJones Just been looking into this, but there are something like 8 different nonces to check. This is going to be a maintenance nightmare, quite apart from the fact that I stand by my point that it's totally useless to do this as - I repeat-:

> The alternative would be to check the nonce + rest and **if ok, not show the notice** and **if not ok, also not show the notice**, so the end result would be exactly the same, just with more code.

PR #514, opened 2016-01-07
Various minor fixes.

jrfnl:
* Stay in line with core: page titles now use H1, not H2.
* Fix incorrect concatenation for translated strings.
* Improved variable names and remove superfluous condition - see #507.
* Use local variable instead of global where available.
* Re-sync the bulk_install() method with the core bulk_upgrade() method.
* Correct and clarify the documentation for the example function.
* Fix two wrong text domain slugs in the example file.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -245,9 +245,9 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 * Adds a reference of this object to $instance, populates default strings,
 		 * does the tgmpa_init action hook, and hooks in the interactions to init.
 		 *
-		 * @internal This method should be `protected`, but as too many TGMPA implementations
+		 * {@internal This method should be `protected`, but as too many TGMPA implementations
 		 * haven't upgraded beyond v2.3.6 yet, this gives backward compatibility issues.
-		 * Reverted back to public for the time being.
+		 * Reverted back to public for the time being.}}
 		 *
 		 * @since 1.0.0
 		 *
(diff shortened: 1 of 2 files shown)


class-tgm-plugin-activation.php
```diff
@@ -3108,6 +3108,7 @@ public function run( $options ) {
 					 *
 					 * @see Plugin_Upgrader::bulk_upgrade()
 					 * @see https://core.trac.wordpress.org/browser/tags/4.2.1/src/wp-admin/includes/class-wp-upgrader.php#L838
+					 * Last synced: Dec 31st 2015 against https://core.trac.wordpress.org/browser/trunk?rev=36134
```
GaryJones: Move non-tagged content above the PHPDoc tags.

jrfnl: Added the `@internal` tag now as the comment does not make sense if moved up.

GaryJones: For an internal note, it should use the inline format. See https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#78-internal

jrfnl: I'll adjust. All the same: the `{}` annotation is a quite recent change and hasn't been in use that long. Secondly the PSR is still in concept phase, so subject to change and as the annotation is a change, it may well change back to how it always used to be used, i.e. without the `{}`.

jrfnl: Adjusted all other occurrences of @internal as well.

GaryJones: phpDocumentor, naturally (same developer), already follows the same format: http://www.phpdoc.org/docs/latest/references/phpdoc/tags/internal.html. 

https://github.com/WordPress/WordPress/blob/master/wp-includes/shortcodes.php#L17-L20 is one instance of the format in WP core, though they are other formats too that have not have been fixed up yet.

PR #516, opened 2016-01-09
Fix i18n functions

ramiy:
In WordPress 4.4 [we fixed many translation functions in the core](https://core.trac.wordpress.org/query?milestone=4.4&component=I18N&reporter=ramiy&groupdesc=1&group=milestone&col=id&col=summary&col=reporter&col=milestone&col=type&col=component&order=priority). The same coding standards should be implemented in TGM too.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -2466,7 +2466,7 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
 		 * @since 2.2.0
 		 */
 		public function no_items() {
-			echo wp_kses_post( esc_html__( 'No plugins to install, update or activate.', 'tgmpa' ) . ' <a href="' . esc_url( self_admin_url() ) . '"> ' . esc_html__( 'Return to the Dashboard', 'tgmpa' ) . '</a>' );
+			echo esc_html__( 'No plugins to install, update or activate.', 'tgmpa' ) . ' <a href="' . esc_url( self_admin_url() ) . '"> ' . esc_html__( 'Return to the Dashboard', 'tgmpa' ) . '</a>';
 			echo '<style type="text/css">#adminmenu .wp-submenu li.current { display: none !important; }</style>';
 		}
 
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1918,7 +1918,13 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 */
 		public function show_tgmpa_version() {
 			echo '<p style="float: right; padding: 0em 1.5em 0.5em 0;"><strong><small>',
-				esc_html( sprintf( _x( 'TGMPA v%s', '%s = version number', 'tgmpa' ), self::TGMPA_VERSION ) ),
+				esc_html(
+					sprintf(
+						/* translators: %s: version number */
+						__( 'TGMPA v%s', 'tgmpa' ),
+						self::TGMPA_VERSION
+					)
+				),
 				'</small></strong></p>';
 		}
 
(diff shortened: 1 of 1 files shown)


jrfnl:
Hi @ramiy Nice one! Thank you.
I'll look through it in detail in a moment.

class-tgm-plugin-activation.php
```diff
@@ -2272,7 +2278,8 @@ protected function get_plugin_status_text( $slug ) {
 			}
 
 			return sprintf(
-				_x( '%1$s, %2$s', '%1$s = install status, %2$s = update status', 'tgmpa' ),
+				/* translators: %1$s: install status, %2$s: update status */
+				__( '%1$s, %2$s', 'tgmpa' ),
 				$install_status,
```
jrfnl: As this 'string' is so generic, should it not have a translators comment *and* a context through _x ?

ramiy: @jrfnl,

The **context** is used to describe the entire string, **translators comment** are used to describe specific placeholders like `%s`, `%1$s`, `%2$s` and others.

Translators comment location

In this case, translators comment are needed because we have two placeholders in this string.

As for the context, I'm not sure. We can set "**Install/Update Status**" as a context, but It's not providing more information than the translator comment.

GaryJones:
I'm +1 for everything happening here. I was aware that `_x` ≠ translator comment, but hadn't had time to go through it, so thank you @ramiy!

class-tgm-plugin-activation.php
jrfnl: In this case, I'd prefer to have both the translators comment and the context. The TGMPA library is included in themes as well as plugins and in the case of wp.org hosted themes, the text-domain is replaced by the theme text-domain.
If, for whatever reason, the theme would have the same generic string in use, things could become very confusing for translators.
For the other strings, I'd think that to be far less likely.

ramiy: It's a very good point. I agree with you.

class-tgm-plugin-activation.php
```diff
@@ -2459,7 +2466,7 @@ public function column_version( $item ) {
 		 * @since 2.2.0
 		 */
 		public function no_items() {
-			printf( wp_kses_post( __( 'No plugins to install, update or activate. <a href="%1$s">Return to the Dashboard</a>', 'tgmpa' ) ), esc_url( self_admin_url() ) );
+			echo wp_kses_post( esc_html__( 'No plugins to install, update or activate.', 'tgmpa' ) . ' <a href="' . esc_url( self_admin_url() ) . '"> ' . esc_html__( 'Return to the Dashboard', 'tgmpa' ) . '</a>' );
```
jrfnl: Considering that the strings are now concatenated and all individual bits are escaped, I think we can get rid of the `wp_kses_post()` warpper.

ramiy: Done.

jrfnl: Awesome! Thanks. If you could squash the commits, I'll merge it.

ramiy: Squash the commits??

jrfnl: As all the changes in the PR are related, it makes more sense, having one commit containing them all, then having four different commits for it.
Squashing is a way to archive that.

See: http://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit

jrfnl: Hi @ramiy I'd really like to merge this. Any chance you can do the squash in the near future ? Thanks!

Let me know if you need any help with that.

ramiy: @jrfnl can't squash this. I use SourceTree for windows, It does not support squash, only the mac version allow the squash.

Does it really matter if you merge one commit or four different commits?

GaryJones: Do it on the command line then: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

ramiy: Can't do it, I have issues with SourceTree terminal:

screenshot

Sorry guys...

jrfnl: Hi @ramiy Sorry to hear you're having trouble with this. The link I send you originally also contains a step by step example of how to do this manually if squash is not supported. Maybe that can help ?

Issue #474, opened 2015-09-03
'tgmpa' text domain


weavertheme:
I've just learned while working with Otto, the Theme Check author, on another issue that they don't consider using multiple text-domains with a theme to be acceptable practice.

The recommendation (and it may become a requirement) is to, at the moment, manually replace all instances of 'tgmpa' text domain strings with the theme's own text domain slug. As before, only a string constant may be used in translation function calls to specify the text domain - and that may need to be the theme text domain slug.

This is not a great choice as 'tgmpa' is used for multiple purposes. 

But if this becomes a new Theme Check requirement, it would be good if it were easier to make the change with a simple search/replace on a string unique to the text domain value.

GaryJones:
Search and replace `, 'tgmpa'`. That's unique to text domains.

weavertheme:
A good thing to know now.

The very latest Theme Check will now flag as an INFO if the theme uses more than one text domain:

"INFO: More than one text-domain is being used in this theme. This means the theme will not be compatible with WordPress.org language packs.
The domains found are weaver-xtreme,  tgmpa".

So while not yet a WARNING, it is likely to lead to questions from other theme developers, and this discussion has the solution!

GaryJones:
Addressing the text domain is on the list. See #218 .

jrfnl:
FYI: The  [Custom TGMPA Generator](http://tgmpluginactivation.com/download/), which is now available on the website, fixes this.

milutinjaric:
No it still has the issue of not changing text domain

nandlal-tjm:
I have same issue

jrfnl:
@DJMIMi @nandlal-tjm The [Custom TGMPA Generator](http://tgmpluginactivation.com/download/) _will_ change the text-domain for you providing you give it the correct information.
If this is not working for you, please open a new topic with a bug report describing what you entered in the generator form and how the received zip file is different from what you expected.

milutinjaric:
New issue opened here https://github.com/TGMPA/TGM-Plugin-Activation/issues/624

Issue #218, opened 2014-08-26
Translation support


cfoellmann:
It is nice that functions like `__()` and `_e()` are being used but using a translation for the library is "complex".
Since I am will need it soon I will have a look at how to do it best. Will post sample code and potentially PRs here.

GaryJones:
I think the reference has changed since you linked to it.

I think there's a wider issue here though. Currently, if TGMPA is included in a theme, then that theme can generate a .pot with the TGMPA strings, but they may not be applied since TGMPA has a PHP variable for a textdomain. If they exclude it, then they don't appear in any .pot file. I think we need to be providing the .pot, .po and .mo (GlotPress install somewhere found via translate.tgmpluginactivation.com) for TGMPA ourselves, and then have the textdomains as `tgmpa` or `tgm-plugin-activation`.

cfoellmann:
A lasting reference is: https://github.com/WebDevStudios/CPT_Core/blob/df1b621bcb03c6a96a13204940600a11b6a62b82/CPT_Core.php#L309-L322

The textdomain is not a var (anymore), right? `__( 'Install Required Plugins', 'tgmpa' )`
I think opening a new GlotPress is not the best way to go. http://wp-translations.org/ might be an option.

jrfnl:
@GaryJones I believe having TGMPA native language files is not the right way to go. Some text string now literally say 'theme' while the library may also be included in plugins, so leaving the providing of the language strings up to the implementing theme/plugin makes a lot of sense.
_**How**_ that is done, however, might need another look, but most of the current implementation of it in the develop branch (with the strings being passed via the config), makes sense to me.
The only problem you may then run into is conflicting translation between multiple implementations in use in the same WP client install.

GaryJones:
If a theme and several plugins in an install all use TGMPA, then that's lots of the same strings that need translating.

If we did it ourselves, providing the localisations as well, then that's a whole bunch of strings that can be excluded from the theme and plugin .pot files. We translate it once, correctly, and load the textdomains from the right location, and those that include that include TGMPA moving forward can include the languages folder too.

GaryJones:
https://ulrich.pogson.ch/future-pot-generators from @grappler also suggests TGMPA having it's own language file(s).

grappler:
I do think that libraries should manage their own translations but it is not always so easy. The time is would work great is if a theme or plugin was using composer to define the dependency and loading all of the files.

[...]

PR #519, opened 2016-01-10
Add a Custom TGMPA Generator to the website.

### LINKED ISSUE #186: several "issues" with 2.4 release
Before version 2.4 the class was using `add_submenu_page`this has now been changed to `add_theme_page`. Although that perhaps makes a lot of sense for themes using the class, it makes no sense whatsoever for plugins using the class.

Also after successful installation of the required plugin it makes no sense (in a plugin scenario) to "return to the dashboard". For starters, the user did not come from the dashboard and I much rather have that the user returns to the Plugins page, because there is a link to my plugin's settings available.

Then I don't understand the switch to the tgmpa text domain. This class is included in themes and plugins, wouldn't it make much more sense that the theme's/plugin's translation files contain the strings of the class?

### LINKED ISSUE #462: Adding class, does not validate WP Theme Check
Hello, I want to upload a theme in wordpress.org and I'm using the Theme check before I upload. 

It shows this Issue:

REQUIRED: class-tgm-plugin-activation.php. Themes should use add_theme_page() for adding admin pages.
Line 609: $this->page_hook = call_user_func( 'add_submenu_page', $args['parent_slug'], $args['page_title'], $args['menu_

Only has to change the function name?

### LINKED ISSUE #475: add_theme_page() Theme Check error

jrfnl:
The Custom TGMPA Generator takes the zip file from the latest release (for now: needs to be uploaded after each release as I can't find a hard link to the zip on GitHub) and adjusts the contents based on settings provided by the end-user in a form in the download page of the website and then serves the customized .zip file to the user.

screenshot

Changes made to the website to support this:
* removed most links to the GH downloads and redirected them to the download page to encourage the use of the form.
* added a number of javascript libraries used by the custom generator which will conditionally load - i.e. only on the download page.
* added octicons font for new download icon
* added the v2.5.2 release zip file to use as base for the generator.
* added a blogpost announcing the custom generator (_**tentative post date January 15 - to be adjusted based on when this will be merged**_)

This PR presumes the related pull requests to the TGMPA library: #520, #521 and #522 will be merged for the 2.6.0 release, but will already function as is for the v 2.5.2 release.

If Theme developers who publish via wp.org use the Generator to create their package, the Theme Check plugin should (will) no longer trigger warnings on the use of TGMPA.

This should also reduce the number of end-user support requests I've come across in the fora caused by too aggressive search-and-replace actions by theme authors on the class-tgm-plugin-activation.php file.

The Generator will apply the following changes to the TGMPA release package:

Adjustment | Affected files | WP.org Theme | ThemeForest Theme | Other Theme | Plugin |
---------- | -------------- | ------------ | ----------------- | ----------- | ------ |
Replace text domain in translation functions |  | in the example file and the class | in the example file only| in the example file only | in the example file only |
Adjust the file include call to the most appropriate one | example file | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
Adjust the file path for the bundled plugin to the most appropriate one | example file | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
Prefix the function name for the function providing TGMPA, including adjusting the hook in | example file | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
Adjust the config id for the notices | example file | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
Adjust menu location & capability config variables | example file | variables removed | variables removed | (no adjustment needed) | :white_check_mark: |
Adjust `add_admin_menu()` function - only have the `add_theme_page()` call in there | class file | :white_check_mark: | :white_check_mark: |  |  |
Remove `load_textdomain` related hook-ins and functions (PR #521) | class file | :white_check_mark: |  |  |  |
Remove `/languages/` directory | | :white_check_mark: | | | |
Remove development related files | | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |

If the file was adjusted, the `@version` tag at the top of the file will also be adjusted to read `@version  x.x.x for plugin/theme Theme Name/Plugin Name` so it will be easy to spot a file which was adjusted using the custom TGMPA generator.
**[Update]** The version tag will now read `@version  x.x.x for plugin/theme Theme Name/Plugin Name for publication on WordPress.org/ThemeForest` if either of those two has been choosen as publication channel.

@Stephen-Cronin Are there any specific requirements for the Themeforest themes using TGMPA which are not covered yet ? If so, please let me know and I'll see what I can do.

/cc @otto42

Resolves #475, resolves #476, resolves #460, resolves #462, resolves #186


_includes/generator-form.html
```diff
@@ -0,0 +1,63 @@
...
+			</fieldset>
+
+			<fieldset>
+			<label for="tgmpa-name">Name:</label>
+			<input type="text" id="tgmpa-name" name="tgmpa-name" placeholder="Theme or Plugin Name" value="" pattern="[A-Za-z0-9 _-]+" title="Allowed characters: alphanumeric, space, underscores and dashes" />
+			</fieldset>
+
+			<fieldset id="tgmpa-form-wporg">
+				<input type="checkbox" id="tgmpa-wporg" name="tgmpa_wporg" value="1">
+				<label for="tgmpa-wporg">Do you intend to distribute the theme via wp.org ?</label>
```
GaryJones: Expand to the full `wordpress.org`, since wp.org redirects to wpbeginner.com.

_includes/generator-form.html
```diff
@@ -0,0 +1,63 @@
+{::options parse_block_html="false" entity_output="as_input" /}
+<div class="tgmpa-generator-form">
+	<form id="generator-form" method="POST">
+		<input type="hidden" name="tgmpa_generate" value="1" />
+
+		<section class="tgmpa-form-primary">
+			<label for="tgmpa-flavor">I'm going to use TGMPA in a:<span class="required">*</span></label>
```
GaryJones: For accessibility, don't make this a label, since it has no direct field, but use a legend inside of the fieldset instead. See http://webaim.org/techniques/forms/controls#radio as an example.

_includes/generator-form.html
```diff
@@ -0,0 +1,63 @@
...
+			<fieldset>
+				<input type="radio" id="tgmpa-flavor-theme" name="tgmpa-flavor" value="parent-theme" required />
```
GaryJones: Use `required="required"` for the XML form of HTML5.

_includes/generator-form.html
```diff
@@ -0,0 +1,63 @@
...
+				<label for="tgmpa-flavor-child-theme">Child Theme</label>
+				<br />
+				<input type="radio" id="tgmpa-flavor-plugin" name="tgmpa-flavor" value="plugin" required />
+				<label for="tgmpa-flavor-plugin">Plugin</label>
+			</fieldset>
+		</section>
+
+		<section class="tgmpa-form-secondary">
+			<fieldset>
+			<label for="tgmpa-text-domain">Text Domain:<span class="required">*</span></label>
```
GaryJones: Consider using the word `required` in the label, so it's read out as such in screen readers. See http://usability.com.au/2013/05/accessible-forms-2-required-fields-and-extra-information/

`Text domain (required):`

_includes/generator-form.html
```diff
@@ -0,0 +1,63 @@
...
+			</fieldset>
+		</section>
+
+		<section class="tgmpa-form-ternary">
+			<div class="generator-form-submit">
+				<input type="submit" class="tgmpa-form-submit" name="tgmpa_generate_submit" value="Generate" />
+			</div>
+	
+			<div id="spinner"></div>
+			<div class="generator-feedback">
```
GaryJones: If this section uses JavaScript to appear, then some use of `aria-live` is needed so screen reader users' attention is drawn to it.

_includes/generator-form.html
```diff
@@ -0,0 +1,63 @@
...
+				<p id="generator-feedback"></p>
+				<p id="report-generator-error">Please try again. If the error persists, please open a <a href="https://github.com/TGMPA/TGM-Plugin-Activation/issues">bug report</a> quoting the error message.</p>
+			</div>
+		</section>
+	</form>
+
+	<div class="generator-notes">
+		<ul>
+			<li>This generator works on Firefox, Chrome, Opera &ge; 15 and IE &ge; 10 (but NOT in compatibility view).</li>
+			<li>No data is saved by us, all the processing is done via javascript in the browser.</li>
```
GaryJones: `javascript` => `JavaScript`.

_includes/header.html
```diff
@@ -61,15 +62,16 @@
...
 
 						<div class="download">
-	  						<a href="{{ site.tgmpa.github }}/archive/{{ site.tgmpa.version }}.zip" class="latest-zip">
-								<img border="0" width="90" alt="Download latest release zipfile" src="{{ site.tgmpa.zipimage }}">
-							</a>
-	  						<a href="{{ site.tgmpa.github }}/archive/{{ site.tgmpa.version }}.tar.gz" class="latest-tar">
-								<img border="0" width="90" alt="Download latest release tarball" src="{{ site.tgmpa.tarimage }}">
-							</a>
+							<p class="download-icon">
+								<a href="{{ '/download/' | prepend: site.tgmpa.url }}" title="Download the latest release">
```
GaryJones: Don't use the `title` attributes - they are all but worthless. Use plain text (may be visually hidden with CSS (not `display: none;`)) with the icon. Relying on an icon alone is not sufficient for screen readers.

_posts/2016-01-15-custom-tgmpa-generator.md
```diff
@@ -0,0 +1,34 @@
...
+
+And then when TGMPA is updated, you have to do it all over again.
+
+<p class="align-right">
+	![screenshot-1]({{ '/images/screenshots/tgmpa-generator-medium.png' | prepend: site.tgmpa.url }})
+</p>
+
+Well, no more. We've heard you and we've worked hard to make this easier for you.
+
+So today, we are releasing an _**Custom TGMPA Generator**_. You can find it on the [Download] page.
```
GaryJones: `an` => `a`.

_posts/2016-01-15-custom-tgmpa-generator.md
```diff
@@ -0,0 +1,34 @@
...
+
+Just tell the form the _text-domain_ of your theme or plugin, what sort of WordPress add-on you will be including TGMPA in and, if it's a theme, whether you intend to publish it on wordpress.org and we'll generate a custom download of the current version of TGMPA for you with all the relevant code already adjusted.
```
GaryJones: "tell the form" is an odd choice of wording. "Fill out the form with" would be better.

css/style.css
```diff
@@ -140,12 +140,28 @@ hgroup.text {
...
 .download a {
+	color: #757575;
 	display: inline-block;
 }
+.download a:hover {
+	color: #5490ca;
+}
+
+.download .mega-octicon {
+	font-size: 112px;
```
GaryJones: Indentation.

css/style.css
```diff
@@ -140,12 +140,28 @@ hgroup.text {
...
+    text-align: center;
+    text-shadow: 2px 2px 5px #666666;
```
GaryJones: Use short form of hex codes where possible.

GaryJones:
Great job!

_includes/generator-form.html
jrfnl: Oh dear, I though that battle had been fought (and won by Automattic). Will change.

css/style.css
jrfnl: Spaces from a copy/paste will get me every time ;-)

css/style.css
jrfnl: The actual site uses the minified version of the css which has the short form. For the dev version of the css, long form is fine.

_includes/generator-form.html
jrfnl: I've added the `aria-required` attribute instead, support when having both the HTML5 as well as the WAI attribute is reasonable. http://john.foliot.ca/required-inputs/

_includes/generator-form.html
jrfnl: Adjusted for that and the wporg fieldset. Hope I got it right.

The wporg fieldset only appears when either parent theme or child theme has been selected, and disappears again if plugin is selected. State on load is hidden.

FYI: Other usability js: - when the user fills in the 'Text domain' field, the 'Function prefix' and 'Name' field will be filled out automatically if they haven't been filled out yet.
Function prefix will be the text domain value with dashes replaced by underscores.
Name will be the text domain value with spaces instead of dashes and written in Title Case.

Validation of all three text fields is done via an HTML5 pattern if the browser supports it and additionally via js on submit.

_includes/header.html
jrfnl: Okidoki, this does mean the icon will be an image, not a font-icon, but I'm ok with that. I will leave in the title as well as for those users who don't recognize the icon and don't use screen readers, it gives a semi-useful mouse-over comment.

jrfnl:
Changes all committed. Let's give @Stephen-Cronin and @otto42 a few days to respond before I squash & we merge.

_includes/header.html
GaryJones: If you've got it in plain text (good), remove it the title attribute. It serves no purpose. For those who have got title attributes being read out in screen readers (not the default), it would read the same thing out twice.

Really, this is all a done deal in the accessibility expertise. Title attributes should not be used except on `iframe` elements.

jrfnl: Ok, removing it is then. My front-end accessbility knowledge is a bit out of date it seems ;-)

> Title attributes should not be used except on iframe elements.

Though this sounds out of date too - title *should* be used on input elements which have the HTML5 `pattern` attribute set as that will allow the browser to display as an error hint.
http://www.w3.org/TR/2011/WD-html5-20110525/common-input-element-attributes.html#the-pattern-attribute

Stephen-Cronin:
Looks exciting! I'll have a look into this and try to get back to you tomorrow.

_includes/header.html
GaryJones: Unless screen readers have an override for reading out title attributes on input elements with pattern attributes, even when their settings say not to read out title attributes, I'd rather see this pattern described within the label, or even placeholder text (i.e. "1ABC" for the spec example), or some other visual description. http://www.deque.com/blog/accessible-client-side-form-validation-html5/ has a few suggestions.

cc @afercia @rrwd

Stephen-Cronin:
I have a question about what happens when the "Do you intend to distribute via wordpress.org" checkbox is selected. Is the only difference that the code allowing the menu item to be moved from Appearance is removed? 

If that's the only difference, then we actually want that for ThemeForest themes as well. They need to pass Theme-Check, including the 'add_submenu_page' check. I'm not sure what that means for the option name - "Do you intend to distribute via wordpress.org or ThemeForest" is a bit clunky.

The only other thing is the text domain issue. From the table above, the text domain won't be replaced in the class for either .org or non .org themes. Have you had any feedback from @otto42 or someone else from the .org Theme Review Team on this? I believe that language packs are coming at some point and that using more than one text domain will not be allowed on .org. Various people have mentioned this in the past (eg #474, #218), but now may be a good time to get an update from official sources on when this may be happening.

As for ThemeForest themes, I know that translations under the tgmpa text domain are on the roadmap for 3.0 and I agree that that's the neatest solution. However, we are still discussing whether we may or may not prefer the text-domain replaced in the class (even though we don't have the technical driver that .org has). I'll try to get that discussion wrapped up asap. Thanks!

jrfnl:
Hi @Stephen-Cronin, thanks for having a look.

> what happens when the "Do you intend to distribute via wordpress.org" checkbox is selected. Is the only difference that the code allowing the menu item to be moved from Appearance is removed? 

No, see the table. (Strange, seems the headers for the columns had gotten mixed up, corrected now - the more extensive changes are for org themes, the minimal ones for plugins)

> The only other thing is the text domain issue. From the table above, the text domain won't be replaced in the class for either .org or non .org themes.

Actually it will be. For non-org themes and for plugins, the text domain will only be replaced in the example file.
For org themes, the text domain will be replaced in *both* the example file as well as the class file.

Similarly, for org themes, the load textdomain call we intend to add, will be removed to prevent conflicts with Theme Check.

Sorry for the confusion.

GaryJones:
> Actually it will be. For non-org themes and for plugins, the text domain will only be replaced in the example file.
> For org themes, the text domain will be replaced in both the example file as well as the class file.

Any reason they couldn't be replaced under both conditions? What benefit is there for keeping `tgmpa` within a theme or plugin with a different text domain?

jrfnl:
> Any reason they couldn't be replaced under both conditions? What benefit is there for keeping tgmpa within a theme or plugin with a different text domain?

No reason why we couldn't replace. The current benefit is that we've already got translations ready for some languages.

Then again, just thinking, you can have two `load_,,_textdomain()` for the same domain in a theme/plugin. The second time, WP will just merge the new translation with the previous one, so we *could* even choose to replace the textdomain in all translation calls in all versions and leave the code that handles the loading of the textdomain in, just point it to the native textdomain.

Not sure how Theme Check deals with two `load_.._textdomain()` calls within one theme though. @otto42 ?

Anyway, this would mean we also wouldn't need to remove the `languages` subdirectory. However, we would need to rename all the .mo files within it to either `{locale}.mo` for themes or `{textdomain}-{locale}.mo` for plugins.
Or maybe even better: instead of the renaming of the files (which may be iffy from within js with it being binary data which I'd be reading out from the zip and writing back under a different name), I can just add a low-level filter to manipulate the file name. Hmm.. actually no, I can't, as I won't know the path the theme or plugin will use for TGMPA, so I won't be able to distinguish between 'our' load_textdomain call and 'theirs'.
So to make this viable, I'd need to test if the renaming of of the binary .mo files from within js will work and won't mangle the files. If I can get it to work, browser support for the Generator will probably be even more limited than it is now.

All together, it's easier not to (replace the textdomain).

Stephen-Cronin:
Hi @jrfnl 

Thanks! That makes a lot more sense now (ie the table column headers being mixed up). :)

We've finished our discussion and decided that we're all good with the 'tgmpa' text domain remaining in the class (if you decide to keep it that is!) and being loaded (when you add that), etc. It's great you are planning to provide translations and it does make more sense to provide them centrally, rather than having thousands of different themes having to replicate them and keep them up to date etc.

However, as mentioned above, we would like the menu location & capability config variables removed and the add_admin_menu() function adjusted to only have add_theme_page(). 

Which means we don't fit either the .org or non .org columns above - hope that doesn't cause too many problems!

jrfnl:
Hi @Stephen-Cronin,

That shouldn't be too hard, let me have a quick look how that'll work out user interface-wise, but the principle of it is easy enough to implement.

I'll ping you when I have updated the branch.

jrfnl:
@Stephen-Cronin All done. I've updated the screenshot & the table to reflect the changes made.

jrfnl:
Ok, made some more adjustments.

If anyone wants to see what the output of the Generator looks like, I've uploaded the 7 different flavours of output for you to peruse: https://adviesen.stackstorage.com/index.php/s/24rI27wv7p60vyr

(will leave them online until this PR is merged)

Stephen-Cronin:
@jrfnl - Thanks. Table and screenshot look great! Will check out the downloads.

Otto42:
Language pack support for all themes and plugins already exists on w.org. We don't currently "enforce" it, but we will eventually. Soft sell while we ramp it up, basically.

Otto42:
Note that we would want themes to use one text domain. Replication of strings is not an issue, since all translations for language packs are centrally managed at translate.w.org. Translate once, everybody gets the benefit.

jrfnl:
> Note that we would want themes to use one text domain. ... Translate once, everybody gets the benefit. 

@Otto42 Thanks for taking the time to have a look at this and confirming we're on the right track as that's exactly what the Generator is about. For those people publishing their themes via wp.org, we're trying to make it as easy as possible to include TGMPA and still use only one text-domain. The Generator will automatically adjust the TGMPA code to match the current requirements for using the language packs as well as sorting out the `add_theme_page()` issue.

For those people not publishing on wp.org (and therefore not getting the benefit of translate.w.org), having TGMPA native translations available will be more efficient.

Maintaining different versions of the same (main) repository for different purposes, makes maintenance really hard. This way, we can still cater to the specific needs of wp.org (and Themeforest), without the need for different branches for different uses/publication channels.

jrfnl:
@GaryJones @Stephen-Cronin @Otto42 Just checking: anything which would still need to change ? If not, I'll squash the commits so we can merge.

jrfnl:
Ok, time to get this baby going I'd say ;-)

Rebased, squashed & blogpost date updated. Ready for merge.

jrfnl:
> If anyone wants to see what the output of the Generator looks like, I've uploaded the 7 different flavours of output for you to peruse: https://adviesen.stackstorage.com/index.php/s/24rI27wv7p60vyr

FYI - I'm taking the examples down now the generator is online. If anyone feels they should stay online for whatever reason, let me know.

PR #521, opened 2016-01-10
Add loading of TGMPA native translation files.

jrfnl:
Implemented in a way that the language loading will work independently of whether TGMPA is included in a (must-use) plugin or a parent/child theme.
Includes compatibility for looking in both the WP_LANG_DIR/plugins/ as well as the WP_LANG_DIR/themes/ directory for translations served by wp.org.

I was hesitant to add this code when I wrote it initially some months ago, but now with the Custom TGMPA Generator, that's no longer an issue.
The PR for the Generator #519 already includes the necessary code to remove the code from this PR from TGMPA for wp.org hosted themes.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -260,6 +260,15 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			// Announce that the class is ready, and pass the object (for advanced use).
 			do_action_ref_array( 'tgmpa_init', array( $this ) );
 
+			/*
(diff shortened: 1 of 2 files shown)


jrfnl:
Added one extra bit of lean loading code: https://github.com/TGMPA/TGM-Plugin-Activation/pull/521/files#diff-6f3dff769e44625dd72b031675006eb3R455 just in case & adjusted the regex in the generator to match.

jrfnl:
Squashed the commits into one.

PR #524, opened 2016-01-11
Add russian language

vburlak:
Add russian translation files with latest update from `tgmpa.pot` file


jrfnl:
Awesome! Thank you.

PR #541, opened 2016-02-22
Escape all translation functions

kisonic:
Escape all translation functions. Recommended for Themeforest.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -328,11 +328,11 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 
 			// Load class strings.
 			$this->strings = array(
-				'page_title'                      => __( 'Install Required Plugins', 'tgmpa' ),
-				'menu_title'                      => __( 'Install Plugins', 'tgmpa' ),
-				'installing'                      => __( 'Installing Plugin: %s', 'tgmpa' ),
-				'updating'                        => __( 'Updating Plugin: %s', 'tgmpa' ),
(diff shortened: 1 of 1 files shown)


jrfnl:
Thank you for your pull request. Unfortunately your PR would result in most strings being double escaped.

Escaping should be done *__late__*, i.e. at the point the string is send to the screen, not earlier. If you'll look at those instances, you'll see that strings are being escaped.

Please feel free to refresh this PR (or send in a new one) if it would turn out that we missed a few.

GaryJones:
> Escaping should be done late, i.e. at the point the string is send to the screen, not earlier.

To clarify further, one can only escape, once you know in which context the string is being used. Is it in markup, JavaScript, textarea, attribute or URL attribute? Hence, late escaping is the best practice.

kisonic:
Yes, late escaping is the best practice. But I don't find escaping even late. For example:

line 331 `'page_title' => __( 'Install Required Plugins', 'tgmpa' ),`
line 688 `'page_title'  => $this->strings['page_title'],`
line 715 `$this->page_hook = call_user_func( 'add_theme_page', $args['page_title'], $args['menu_title'], $args['capability'], $args['menu_slug'], $args['function'] );`

Or I miss something?

GaryJones:
The `page_title` is passed through to WP globals `$menu / $submenu`, via the `add_theme_page()` function, to building out the admin menu. WP then the page title available via `get_admin_page_title()`, and we [escape](https://github.com/TGMPA/TGM-Plugin-Activation/blob/develop/class-tgm-plugin-activation.php#L746) that.

kisonic:
Ok, I missed it. Thanks for the explanation.

PR #547, opened 2016-03-21
Minor tweaks

jrfnl:
* See if we can make the build faster by asking travis to cache the Gems bundle
* Better no markdown syntax for HTML form
* Only add robots meta tag if page is published and in sitemap


GaryJones:
Failed build - appears to be checking the syntax of Gems...

jrfnl:
Hmm.. interesting side-effect of enabling bundle caching...

I've now excluded the vendor directory explicitly and the build now passes :sunglasses:

PR #549, opened 2016-03-30
Translate all strings

emiluzelac:
Since they are not a hardcoded, all names should be translatable. 

Example that most ThemeForest and WordPress.org authors will copy as-is and we require them anyways.

### DIFF
--- a/example.php
+++ b/example.php
@@ -61,7 +61,7 @@ function my_theme_register_required_plugins() {
 
 		// This is an example of how to include a plugin bundled with a theme.
 		array(
-			'name'               => 'TGM Example Plugin', // The plugin name.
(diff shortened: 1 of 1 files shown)


jrfnl:
Hi @emiluzelac Thanks for this PR. I'm a bit hesitant to merge it at this point in time as it can easily lead to inconsistent translations of plugin names as theme and plugins would provide them themselves.
The alternative would be to use the translations provided by the plugins themselves, however that would only work for plugins which are already installed and would require us to load the textdomains for all plugins involved on the TGMPA page (like WP core does on the plugins page).

Let me think about this a bit more and please let me know if you have more thoughts on this too.

emiluzelac:
Hey @jrfnl sounds great. Also see: https://github.com/Automattic/theme-tools/pull/16/files I kind of did this at the same time ;)

PR #561, opened 2016-04-19
Added Dutch translations

geoclaps:
Sorry it took a while, I deleted my previous translated files.


jrfnl:
Thanks for taking the time to re-translate! It's much appreciated!

geoclaps:
Not a problem. You're welcome!

PR #563, opened 2016-04-19
Add translator comments to all strings with variables.

jrfnl:
In line with how these are in core.
Includes updated .pot file.


GaryJones:
Yes, yes, yes :-)

PR #565, opened 2016-04-21
Update fr_FR translation

hedii:
Only new fields where added


jrfnl:
Awesome! And thank you for sending the pull in against the feature branch! Perfect!

PR #566, opened 2016-04-21
Update ru_RU translation

vburlak:
Update `ru_RU` translation files with few new strings and add new plural strings


jrfnl:
Awesome! Thank you!

PR #232, opened 2014-10-03
Issue #231: Added notifications when plugins are force activated

creativecoder:
I've added this with the goal of providing feedback to users that indicates why plugins are being automatically reactivated.

Implemented by adding a new method, for possibly extending this to work on forced deactivation, at some point in the future.

Issue #231, opened 2014-10-03
Display an admin notice when a plugin is automatically re-activated


creativecoder:
Currently, when a plugin has force_activation set to true, it will be automatically reactivated by the class, if deactivated from the plugins page in wp-admin.

The notice that appears is still "Plugin deactivated". I think it would be a good idea to remove this notice and replace it with a notice that the plugin has been automatically activated, because it's required by the current theme.

PR #569, opened 2016-04-23
Issue #231: Added notifications when plugins are force activated

creativecoder:
I've added this with the goal of providing feedback to users that indicates why plugins are being automatically reactivated.

Updated from #232


class-tgm-plugin-activation.php
```diff
@@ -371,6 +380,9 @@ public function init() {
 					'The following recommended plugins are currently inactive: %1$s.',
 					'tgmpa'
 				),
+				'notice_force_activation'        => _n_noop(
+					'The following plugin has been automatically activated because it is required by the current theme: %s', 'The following plugins have been automatically activated because they are required by the current theme: %s'
```
jrfnl: This string will also need to be added to the `example.php` file for potential overruling.

class-tgm-plugin-activation.php
```diff
@@ -1124,11 +1136,15 @@ public function notices() {
 			$total_required_action_count = 0;
 
 			foreach ( $this->plugins as $slug => $plugin ) {
-				if ( $this->is_plugin_active( $slug ) && false === $this->does_plugin_have_update( $slug ) ) {
+				$force_activated = in_array( $slug, $this->force_activated_plugins );
```
jrfnl: Please use strict checking in `in_array()`.

jrfnl:
Hi @creativecoder Thanks ever so much for updating the PR! I've added some inline notes about the code itself.

One thing I'm wondering about is whether the 'generic' TGMPA notice is the best place to advise a user about a force activated plugin. Or should a force activated plugin have it's own admin notice ?

IMHO I think splitting them off would be the more user friendly. The generic `TGMPA` notice is basically a _call to action_ and the notice you've now added is a _FYI_.

Also, I can imagine that the notice should only be shown to admin users (who may have tried to deactivate the plugin) and not to the wider range of users who get to see the call to action notice.

What are your thoughts on this ? @GaryJones want to weight in as well ?

GaryJones:
I'm in agreement. It's a great feature, but it should be a separate notice, shown to just the relevant group of users.

class-tgm-plugin-activation.php
```diff
@@ -2023,6 +2039,7 @@ public function force_activation() {
 					} elseif ( $this->can_plugin_activate( $slug ) ) {
 						// There we go, activate the plugin.
 						activate_plugin( $plugin['file_path'] );
+						$this->force_activated_plugins[] = $slug;
```
jrfnl: I'm thinking a transient should be set with this information. The situation - a plugin being force (re-) activated - could come up during a bulk-install which uses frames which would prevent the message from becoming visible. Similarly, another program could with code try to silently deactivate a plugin.

So, what about setting a transient with the slugs of force-activated plugins whenever this occurs ? Then having an admin_notice action which checks if such a transient exists, that we're in the WP back-end and that the current user can activate activate plugins and if so, shows a one-time notice to that user about the plugin(s) which were force activated ?

creativecoder: Thanks for the thoughtful feedback. I like all of these ideas, both from
a UX and engineering perspective. I'm aiming to get to these changes and
additions in the next week or two and will update the pull request.

jrfnl:
@creativecoder Just a quick question - we're about ready to release TGMPA v2.6.0 and I'm wondering whether we should hold off another week or so to include this PR or not. What do you think ?

creativecoder:
I can work on updating the pull request this weekend with specified
changes, if you want to hold back the release until next week, so it can
be included.

jrfnl:
@creativecoder OK, will hold off until after the weekend.
Reminder to self:  We may need to ping the translators (again) for the extra string as well.

creativecoder:
@jrfnl @GaryJones Sorry for the delay! I've update my pull request, as suggested. Please let me know if you see any issues that need to be address before merging.

class-tgm-plugin-activation.php
```diff
@@ -371,6 +371,9 @@ public function init() {
 					'The following recommended plugins are currently inactive: %1$s.',
 					'tgmpa'
 				),
+				'notice_force_activation'         => _n_noop(
+					'The following plugin has been automatically activated because it is required by the current theme: %s', 'The following plugins have been automatically activated because they are required by the current theme: %s'
+				),
```
jrfnl: `_n_noop()` needs the text-domain as the last argument.

example.php
```diff
@@ -177,6 +177,9 @@ function my_theme_register_required_plugins() {
 				'The following recommended plugins are currently inactive: %1$s.',
 				'theme-slug'
 			),
+			'notice_force_activation'         => _n_noop(
+				'The following plugin has been automatically activated because it is required by the current theme: %s', 'The following plugins have been automatically activated because they are required by the current theme: %s'
+			),
```
jrfnl: `_n_noop()` needs the text-domain as the last argument. For the example file this is `theme-slug`.

jrfnl:
Hi @creativecoder I'd like to take the time to properly look at what you've done now and as we've already started the process now of releasing v 2.6.0, I'm inclined to not rush this through and leave it for v2.6.1 if you don't mind.
Added two small remarks so far, but will try and look your PR over properly & test it over the next week or so.

class-tgm-plugin-activation.php
```diff
@@ -448,6 +451,11 @@ public function init() {
 			if ( true === $this->has_forced_deactivation ) {
 				add_action( 'switch_theme', array( $this, 'force_deactivation' ) );
 			}
+
+			// Display forced activation notice, if present.
+			if ( current_user_can( 'manage_options' ) && is_admin() ) {
```
jrfnl: No need for the action if there are no force activated plugins
`if ( true === $this->has_forced_activation && ...`

class-tgm-plugin-activation.php
```diff
@@ -1337,6 +1362,22 @@ protected function display_settings_errors() {
 		}
 
 		/**
+		 * Display admin notice if plugins have been force activated.
+		 *
+		 * @since 2.6.0
+		 */
+		public function display_forced_activation_notice() {
+			if ( $force_activated_plugins = get_transient( 'tgmpa_force_activated_plugins' ) ) {
```
jrfnl: Please don't assign to a variable within a condition. I know the WP project does this regularly, but it's bad practice.

class-tgm-plugin-activation.php
```diff
@@ -1337,6 +1362,22 @@ protected function display_settings_errors() {
...
+				?>
+				<div class="notice <?php echo esc_attr( $this->get_admin_notice_class() ); ?> is-dismissible">
```
jrfnl: No need to go in and out of PHP several times. Just gather the variables and echo message out once you've collected the parts using comma's between arguments.

Also, while the transient should never be empty, an extra defensive coding check to verify this before building the message would not go amiss.

class-tgm-plugin-activation.php
```diff
@@ -2023,12 +2064,31 @@ public function force_activation() {
...
 		/**
+		 * Set/update a transient, appending the new value if the transient already exists.
+		 *
+		 * @since 2.6.0
+		 *
+		 * @param string       $transient_key Name of transient.
+		 * @param string|array $data To set or add to transient.
+		 */
+		protected function append_transient( $transient_key, $data ) {
+			if ( $transient_data = get_transient( $transient_key ) ) {
```
jrfnl: Please don't assign to a variable within a condition.

class-tgm-plugin-activation.php
```diff
@@ -2023,12 +2064,31 @@ public function force_activation() {
...
+				(array) $transient_data[] = $data;
+			} else {
+				$transient_data = (array) $data;
+			}
+
+			set_transient( $transient_key, $transient_data );
+		}
```
jrfnl: Just wondering if we can make this slightly more efficient - if there are five plugins being force activated, the transient (and the associated cache) would be (re)set five times or am I missing something ? Could we collect the values and set the transient in one go ? or would that become very complicated ?

class-tgm-plugin-activation.php
```diff
@@ -1337,6 +1362,22 @@ protected function display_settings_errors() {
 		}
 
 		/**
+		 * Display admin notice if plugins have been force activated.
+		 *
+		 * @since 2.6.0
+		 */
+		public function display_forced_activation_notice() {
+			if ( $force_activated_plugins = get_transient( 'tgmpa_force_activated_plugins' ) ) {
+				?>
```
jrfnl: Re: my edited remark on line 456 - while unlikely, there may be a lingering transient which hasn't been displayed yet for a plugin which in the mean time has been deactived again ? Do we need to check if the force activated plugins we're showing the notice for are still activated before showing them in the notice ?

jrfnl:
@creativecoder Very much like what you've done now! Added some thoughts to the code, but they are mainly about the details, not the principle of it. This is shaping up to be a great PR!

jrfnl:
@GaryJones Quick question - looks like Scrutinizer is not running for PRs from external forks... any ideas ?

class-tgm-plugin-activation.php
```diff
@@ -1236,6 +1218,49 @@ public function notices() {
 		}
 
 		/**
+		 * Build a message string that specifies what actions are needed and the plugins that need those actions.
+		 *
+		 * @since 2.6.0
+		 *
+		 * @param  array  $message Associative array of actions: each key is an action and each value is an array of plugin slugs that need that action.
```
jrfnl: Sorry to be a pain, but the parameter comments should line up, so the text `Associative..` needs some more spaces before it.

class-tgm-plugin-activation.php
```diff
@@ -2023,12 +2064,31 @@ public function force_activation() {
...
 			}
 		}
 
 		/**
+		 * Set/update a transient, appending the new value if the transient already exists.
+		 *
+		 * @since 2.6.0
+		 *
+		 * @param string       $transient_key Name of transient.
+		 * @param string|array $data To set or add to transient.
```
jrfnl: Sorry to be a pain, but the parameter comments should line up, so the text `To set..` needs some more spaces before it.

class-tgm-plugin-activation.php
GaryJones: I've got no objection to this. By going out of PHP, you then benefit from syntax-highlighted markup, and it's the first step to moving markup out of the PHP logic and into view files.

What I'd like to see fixed is changing `esc_attr()` to `sanitize_html_class()` here.

GaryJones:
> Quick question - looks like Scrutinizer is not running for PRs from external forks... any ideas ?

No idea. Didn't you recently update the config for that?

GaryJones:
The `@since` values will need updating - perhaps put them as `2.x.x`, so they'll be easier to find amongst the real `2.6.0` values later on?

class-tgm-plugin-activation.php
GaryJones: Also, only a single space is needed after `@param`.

Yes, that means it won't line-up with the type in the `@return` string, but that's what WP does.

class-tgm-plugin-activation.php
```diff
@@ -1236,6 +1218,49 @@ public function notices() {
...
+		 * @since 2.6.0
+		 *
+		 * @param  array  $message Associative array of actions: each key is an action and each value is an array of plugin slugs that need that action.
+		 * @param  string $line_template String that sprintf is applied to in creating a line of html markup for each action.
+		 * @return string Message
+		 */
+		protected function build_message( $message, $line_template = '%s' ) {
+			$message_output = '';
+
+			krsort( $message ); // Sort messages.
```
GaryJones: The comment is redundant.

class-tgm-plugin-activation.php
```diff
@@ -1236,6 +1218,49 @@ public function notices() {
 		}
 
 		/**
+		 * Build a message string that specifies what actions are needed and the plugins that need those actions.
+		 *
+		 * @since 2.6.0
+		 *
+		 * @param  array  $message Associative array of actions: each key is an action and each value is an array of plugin slugs that need that action.
+		 * @param  string $line_template String that sprintf is applied to in creating a line of html markup for each action.
+		 * @return string Message
```
GaryJones: Full stop after `@return` (and `@param`) descriptions please.

class-tgm-plugin-activation.php
```diff
@@ -2023,12 +2064,31 @@ public function force_activation() {
...
+		 * Set/update a transient, appending the new value if the transient already exists.
+		 *
+		 * @since 2.6.0
+		 *
+		 * @param string       $transient_key Name of transient.
+		 * @param string|array $data To set or add to transient.
+		 */
+		protected function append_transient( $transient_key, $data ) {
+			if ( $transient_data = get_transient( $transient_key ) ) {
+				(array) $transient_data[] = $data;
```
GaryJones: This casting is redundant.

jrfnl: Actually it isn't, but it is not done correctly in this case ;-)
`get_transient()` might return false if the transient is not found or any type of value if it is found. Making sure we don't get a PHP notice here is a good thing, just should be implemented slightly differently, more along the lines of below.

```php
$transient_data = get_transient( $transient_key );
if ( false === $transient_data ) ) {
	$transient_data = array();
}
if (  ! is_array( $transient_data ) ) {
	$transient_data = (array) $transient_data;
}

if ( ! empty( $data ) ) {
	if ( is_string( $data ) ) {
		$transient_data[] = $data;
	} elseif ( is_array( $data ) ) {
		$transient_data = $transient_data + $data;
	}
}
```

class-tgm-plugin-activation.php
jrfnl: The sanitization is already done within the `get_admin_notice_class()` method. `esc_attr()` is correct in this case.

jrfnl:
> No idea. Didn't you recently update the config for that?

I did, but that shouldn't have any effect on when it's run, just what checks it runs.
If I look at the Scrutinizer site, it looks like it does actually run the checks on PRs, but GH does not show them in the check bit... weird

class-tgm-plugin-activation.php
creativecoder: Right... the property `has_forced_activation` indicates if there are plugins that need to be force activated, where the the transient and message are displaying only if they have _already been_ force activated, but we still need to notify the user.

class-tgm-plugin-activation.php
creativecoder: I think the answer to that is whether or not plugins could be deactivated over multiple requests... if yes, then we should keep as is so that the list could aggregate in the transient until it is finally displayed by the admin_notices hook. If not (we know all plugins will be deactivated on the same request), then this could be refactored to collect the plugin list and add to the transient all at once.

class-tgm-plugin-activation.php
creativecoder: Seems like a good idea to me--I've refactored accordingly.

creativecoder:
Thanks for the feedback. I've responded and made several updates. Let me know if you see anything outstanding.

PR #570, opened 2016-04-26
Added Czech localization

vyskoczilova:
Created a new translation for Czech. Credits to Karolína Vyskočilová (native speaker), http://kybernaut.cz .


jrfnl:
Thank you!

PR #571, opened 2016-04-29
More comprehensive scrutinizer config.

jrfnl:
Prevents builds in forks throwing different results (Scutinizer uses a config cascading merge).

Includes minor code changes based on Scrutinizer feedback.
Also reset the Scrutinizer filtered issues and only re-filtered persistent issues we can't currently get rid off for varying reasons.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1015,11 +1015,11 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				$subdir_name = untrailingslashit( str_replace( trailingslashit( $remote_source ), '', $source ) );
 
 				if ( ! empty( $subdir_name ) && $subdir_name !== $desired_slug ) {
-					$from = untrailingslashit( $source );
-					$to   = trailingslashit( $remote_source ) . $desired_slug;
+					$from_path = untrailingslashit( $source );
+					$to_path   = trailingslashit( $remote_source ) . $desired_slug;
 
-					if ( true === $GLOBALS['wp_filesystem']->move( $from, $to ) ) {
-						return trailingslashit( $to );
+					if ( true === $GLOBALS['wp_filesystem']->move( $from_path, $to_path ) ) {
+						return trailingslashit( $to_path );
 					} else {
 						return new WP_Error( 'rename_failed', esc_html__( 'The remote plugin package does not contain a folder with the desired slug and renaming did not work.', 'tgmpa' ) . ' ' . esc_html__( 'Please contact the plugin provider and ask them to package their plugin according to the WordPress guidelines.', 'tgmpa' ), array( 'found' => $subdir_name, 'expected' => $desired_slug ) );
 					}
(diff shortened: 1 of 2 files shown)


class-tgm-plugin-activation.php
```diff
@@ -2713,7 +2713,7 @@ protected function get_row_actions( $item ) {
 			}
 
 			$prefix = ( defined( 'WP_NETWORK_ADMIN' ) && WP_NETWORK_ADMIN ) ? 'network_admin_' : '';
-			return apply_filters( "tgmpa_{$prefix}plugin_action_links", array_filter( $action_links ), $item['slug'], $item, $this->view_context );
+			return apply_filters( 'tgmpa_' . $prefix . 'plugin_action_links', array_filter( $action_links ), $item['slug'], $item, $this->view_context );
```
GaryJones: AFAIK, filters (and the action hook below) need to be named with interpolation, for WP-Parser to be able to read the single string. Unless WP-Parser has been updated to read the first arg token, instead of the first string token...

GaryJones:
Probably should have been separate commits - the config file is fine to merge, but one small query on one of the fixes means the PR can't be merged immediately.

class-tgm-plugin-activation.php
jrfnl: Inetresting. I didn't know that. Bit of a shame that they're promoting bad practices that way, but fair enough.
I'll try and figure out the rule which triggered this change and turn it off in the config.

jrfnl:
See my above comment - exactly the reason why I put in one PR (though I can easily make it two commits of course).

class-tgm-plugin-activation.php
jrfnl: Also just wondering: we're not using the WP-Parser, so why should we be led by what it can parse ?

GaryJones: While we're not, there may be plugins and themes that are using it, so TGMPA shouldn't be the cause of them getting duff auto-generated documentation because it tripped up in some way.

Of course, we could use WP-Parser if we wanted to create some technical API documentation.

jrfnl: Fair point.

> Of course, we could use WP-Parser if we wanted to create some technical API documentation.

I was considering something like that at some point once all the hooks are documented. 
All the same, if the WP Parser is that finicky, we may want to consider using [WP Hook documentor](https://github.com/jrfnl/wp-hook-documentor) instead. At least it parses all variants of hooks ;-)

Or maybe better: fix WP Parser ... (yet another project to fix... *sigh*)

GaryJones:
Are you planning to change these hook names back or not?

jrfnl:
Yes I am, just need to figure out the scrutinizer rule to invert.

jrfnl:
Adjusted & turned into two separate commits, one for the config, one for the code adjustments.

The issue is triggered by the [WPCS rules](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/search?utf8=%E2%9C%93&q=Squiz.Strings.DoubleQuoteUsage.ContainsVar), so no specific Scrutinizer rule I can disable for that. In the WPCS config the severity is `0`, but clearly Scrutinizer cannot deal with this properly, so ended up filtering out the issue in Scrutinizer itself.

jrfnl:
Hang on - some alignment issues were introduced by the translators comments, let me fix those too.

jrfnl:
Ok, all done.

PR #591, opened 2016-05-19
Release-prep version 2.6.1.

jrfnl:
- Upped version numbers
- Added changelog
- Updated .pot file and synced all translation files (and where necessary adjusted the offending string.

As the string replacement issue is very visible, I think it's prudent to do a quick patch-release.
I've prepped a PR for the `gh-pages` branch as well, just need to add the zip file once this is merged.


GaryJones:
This should be `2.6.1`, not `2.6.0.1`. It's a patch release for 2.6.0. Standard semver. The fact it only contains one fix is not important.

jrfnl:
Adjusted.

So when should the x.x.0.1 version numbering be used ?

GaryJones:
> So when should the x.x.0.1 version numbering be used ?

If you must, then as a build / compile version (i.e. internal release), but that's not typically needed in the code we write.

jrfnl:
Shall I merge `develop` into `master` & tag the release ?

GaryJones:
Yes please.

jrfnl:
Done - will send in the website PR in a moment.

PR #593, opened 2016-05-24
Fixed: Use ABSPATH check instead of function check (default way)

Ninos:


jrfnl:
Hi Ninos, could you please split this PR and only have changes to the TGMPA library here ?

The [example plugin has it's own repository](https://github.com/TGMPA/tgm-example-plugin) you can send PRs to. We'll update the zip on a next release.

Ninos:
Sure, I'll split it now :)

PR #594, opened 2016-05-24
Enhancement: Exit if class file is loaded directly (ABSPATH check)

Ninos:
See also #593

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -32,6 +32,11 @@
 	Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
+// Exit if accessed directly
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
 if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 
 	/**
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -32,7 +32,7 @@
 	Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 */
 
-// Exit if accessed directly
+// Exit if accessed directly.
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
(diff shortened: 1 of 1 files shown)


jrfnl:
I'm wondering if adding this will impact people who load TGMPA via a Composer auto-load file. I don't use Composer with this myself, so am really not sure, but seem to remember some issues elsewhere with this kind of code in combination with Composer.

Anyone who is more knowledgable about this want to chip in ?

GaryJones (approved):

widoz:
Should a php file take care of a server problem? Isn't this a server territory?

Ninos:
@widoz no. Sometimes a server should be able to run an alternative file instead of an index.php (e.g. ajax.php / admin.php) in a subdirectory.

widoz:
@Ninos The problem is that a file should not implement side effect and not must take care of wrong server configuration. That my though. 

> Sometimes a server should be able to run an alternative file instead of an index.php (e.g. ajax.php / admin.php) in a subdirectory.

How that is related with the directly access to the file?

GaryJones:
> The problem is that a file should not implement side effect

I completely agree, but, in terms of backwards compatibility, our main file does (it runs a conditional including WP functions in the global scope), and so in this case, we make the file only do anything if its loaded in the context of WP also being loaded.

I was hitting this issue, as I was running unit tests against a theme which included this library file, via Composer, so the tests autoloading (`vendor/autoload.php`) the theme source files was also loading this TGMPA file, and it was throwing a fatal error on the `did_action()` call in the global scope.

Version 3, should we get that far, would aim to have individual classes in their own files, with no side effects, with a single bootstrap file holding the global scope code.

Ninos:
@widoz 
> How that is related with the directly access to the file?

Well, the ABSPATH check is for checking if wordpress was initialized before. So a directly access to the file will not run any code anymore (you'll never get errors by direct access).
I agree with you that in general wp-devs should more code object-oriented and such problems must be fixed by wp-core itself (e.g. running an init-function in the main class instead of just including a file (functions.php for themes & my-plugin.php for plugins).

PR #595, opened 2016-05-25
Update class-tgm-plugin-activation.php

Vishal-Deshpande:
' __ ' Replaced with 'esc_html__( '


jrfnl:
Closing as duplicate of #541 for the same reason.

> Thank you for your pull request. Unfortunately your PR would result in most strings being double escaped.
> 
> Escaping should be done late, i.e. at the point the string is send to the screen, not earlier. If you'll look at those instances, you'll see that strings are being escaped.
> 
> Please feel free to refresh this PR (or send in a new one) if it would turn out that we missed a few.

PR #598, opened 2016-05-27
Fix load textdomain

someden:


GaryJones:
Thanks @kisonic . Could you please split the fixing of the textdomain loading, and the RU strings into different pull requests (branches) please? They are really fixing different issues.

Also, please take note of the code standards issue outlined [here](https://travis-ci.org/TGMPA/TGM-Plugin-Activation/jobs/133387072). Basically, use single-quotes for PHP strings when possible.

someden:
Ok, I will fix it

PR #626, opened 2016-09-14
Red Border for Required Plugins

cyberwani:
added left-border in plugin list for required plugins.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -2100,8 +2100,6 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		 * Adds css to admin head
 		 *
 		 * @since 2.6.2
-		 *
-		 * @return none
 		 */
 		function tgm_css() {
 			echo '<style>
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -2096,6 +2096,13 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			return self::$instance;
 		}
 
+		/**
+		 * Adds css to admin head
+		 *
+		 * @since 2.6.2
+		 *
+		 * @return none
+		 */
 		function tgm_css() {
 			echo '<style>
 			#tgmpa-plugins #the-list .required > th {
@@ -2751,7 +2758,7 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
 		 * @param object $item The current item.
 		 */
 		public function single_row( $item ) {
-			echo '<tr class="'.strtolower($item['type']).'">';
+			echo '<tr class="' . esc_attr( strtolower( $item['type'] ) ) . '">';
 			parent::single_row_columns( $item );
 			echo '</tr>';
 
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -452,7 +452,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				add_action( 'switch_theme', array( $this, 'force_deactivation' ) );
 			}
 
-			add_action( 'admin_head',  array( $this, 'tgm_css' ), 21 );
+			add_action( 'admin_head',  array( $this, 'admin_css' ) );
 		}
 
 		/**
@@ -2097,13 +2097,13 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		}
 
 		/**
-		 * Adds css to admin head
+		 * Adds CSS to admin head.
 		 *
 		 * @since 2.6.2
 		 */
-		function tgm_css() {
+		function admin_css() {
 			echo '<style>
-			#tgmpa-plugins #the-list .required > th {
+			#tgmpa-plugins .tgmpa-type-required > th {
 			border-left: 3px solid #dc3232;
 			}
 			</style>';
@@ -2756,7 +2756,7 @@ if ( ! class_exists( 'TGMPA_List_Table' ) ) {
 		 * @param object $item The current item.
 		 */
 		public function single_row( $item ) {
-			echo '<tr class="' . esc_attr( strtolower( $item['type'] ) ) . '">';
+			echo '<tr class="' . esc_attr( 'tgmpa-type-' . strtolower( $item['type'] ) ) . '">';
 			parent::single_row_columns( $item );
 			echo '</tr>';
 
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -452,6 +452,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				add_action( 'switch_theme', array( $this, 'force_deactivation' ) );
 			}
 
+			// Add CSS for the TGMPA admin page.
 			add_action( 'admin_head',  array( $this, 'admin_css' ) );
 		}
 
@@ -2082,6 +2083,24 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 		}
 
 		/**
+		 * Adds CSS to admin head.
+		 *
+		 * @since 2.6.2
+		 */
+		function admin_css() {
+			if ( ! $this->is_tgmpa_page() ) {
+				return;
+			}
+
+			echo '
+			<style>
+			#tgmpa-plugins .tgmpa-type-required > th {
+				border-left: 3px solid #dc3232;
+			}
+			</style>';
+		}
+
+		/**
 		 * Returns the singleton instance of the class.
 		 *
 		 * @since 2.4.0
@@ -2094,19 +2113,6 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 			}
 
 			return self::$instance;
-		}
-
-		/**
-		 * Adds CSS to admin head.
-		 *
-		 * @since 2.6.2
-		 */
-		function admin_css() {
-			echo '<style>
-			#tgmpa-plugins .tgmpa-type-required > th {
-			border-left: 3px solid #dc3232;
-			}
-			</style>';
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


GaryJones:
Thanks @cyberwani.

Can you please show a screenshot with your change?
Have you checked what happens with the border when the screen is narrow and lines wrap around?
Is the red the same hex code of the attention grabbing red in WP core? (For which there's a Trac ticket open by @afercia) to consolidate these down).

You'll need to ensure you meet code standards - see https://travis-ci.org/TGMPA/TGM-Plugin-Activation/jobs/159883749 for issues with your PR.

cyberwani:
Here are some screenshots:
320px
480px
768px
1440px Cropped

I used same color code used in WordPress warnings.

GaryJones:
Thanks for providing those. Visually, I'm OK with that. @jrfnl ?

@cyberwani The code will still need fixing for code standards.

jrfnl:
Visually I like the change, nice one!

Code-wise, quite apart from the code style issues, I'd like to properly review it before merging and it may be a little while before I have the time.

cyberwani:
@GaryJones @jrfnl I am not familiar with Travis. So, can you both suggest me the further steps?

jrfnl:
Travis is an automated code check which runs on the code and causing one of the red crosses below.
If you click on "Details", you can see which checks are run and which is failing.
In this case it is failing as the code does not comply with the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
For details of what you need to fix, click on the failing build to see the feedback given there: 
https://travis-ci.org/TGMPA/TGM-Plugin-Activation/jobs/159883749#L202

GaryJones:
For reference, the following code standards (formatting of the code) need to be addressed:

```
----------------------------------------------------------------------
FOUND 6 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 2099 | ERROR | [ ] Missing function doc comment
      |       |     (Squiz.Commenting.FunctionComment.Missing)
 2754 | ERROR | [x] Concat operator must be surrounded by a single
      |       |     space
      |       |     (Squiz.Strings.ConcatenationSpacing.PaddingFound)
 2754 | ERROR | [ ] Expected next thing to be an escaping function (see
      |       |     Codex for 'Data Validation'), not 'strtolower'
      |       |     (WordPress.XSS.EscapeOutput.OutputNotEscaped)
 2754 | ERROR | [x] Expected 1 spaces after opening bracket; 0 found
      |       |     (PEAR.Functions.FunctionCallSignature.SpaceAfterOpenBracket)
 2754 | ERROR | [x] Expected 1 spaces before closing bracket; 0 found
      |       |     (PEAR.Functions.FunctionCallSignature.SpaceBeforeCloseBracket)
 2754 | ERROR | [x] Concat operator must be surrounded by a single
      |       |     space
      |       |     (Squiz.Strings.ConcatenationSpacing.PaddingFound)
----------------------------------------------------------------------
```

So, for the indicated lines:
1. Add a full DocBlock in the correct format for the new function you added.
2. Add a space either side of the `.`.
3. Add `esc_attr()` call around the `strtolower()` call.
4, 5 and 6. Add extra spaces as indicated.

class-tgm-plugin-activation.php
```diff
@@ -2096,6 +2096,13 @@ public static function get_instance() {
 			return self::$instance;
 		}
 
+		/**
+		 * Adds css to admin head
```
GaryJones: "CSS" in capitals please.
Add a full stop at the end of the description.

class-tgm-plugin-activation.php
```diff
@@ -2096,6 +2096,13 @@ public static function get_instance() {
...
+		 *
+		 * @since 2.6.2
+		 *
+		 * @return none
```
GaryJones: Leave this `@return` line (and the empty line above it) out please.

class-tgm-plugin-activation.php
```diff
@@ -2093,6 +2095,19 @@ public static function get_instance() {
 
 			return self::$instance;
 		}
+
+		/**
+		 * Adds css to admin head
+		 *
+		 * @since 2.6.2
+		 */
+		function tgm_css() {
```
GaryJones: I'd like this to be renamed to `admin_css` (and the hook reference), just to be a little more specific.

class-tgm-plugin-activation.php
```diff
@@ -451,6 +451,8 @@ public function init() {
 			if ( true === $this->has_forced_deactivation ) {
 				add_action( 'switch_theme', array( $this, 'force_deactivation' ) );
 			}
+
+			add_action( 'admin_head',  array( $this, 'tgm_css' ), 21 );
```
GaryJones: Where has this magic `21` come from? Maybe a comment is needed to explain why that priority was chosen.

class-tgm-plugin-activation.php
```diff
@@ -2093,6 +2095,19 @@ public static function get_instance() {
...
 		}
+
+		/**
+		 * Adds css to admin head
+		 *
+		 * @since 2.6.2
+		 */
+		function tgm_css() {
+			echo '<style>
+			#tgmpa-plugins #the-list .required > th {
```
GaryJones: Do we really need two IDs in the selector? At most, one should be needed, but ideally, none if classes are available.

class-tgm-plugin-activation.php
```diff
@@ -2741,7 +2756,9 @@ protected function get_row_actions( $item ) {
 		 * @param object $item The current item.
 		 */
 		public function single_row( $item ) {
-			parent::single_row( $item );
+			echo '<tr class="' . esc_attr( strtolower( $item['type'] ) ) . '">';
```
GaryJones: This `required` class is too generic - it may well clash with other plugins or maybe even core at some point in the future. Let's go for something like:

```
echo '<tr class="' . esc_attr( 'tgmpa-type-' . strtolower( $item['type'] ) ) . '">';
```

...and updating the CSS selector to match.
(With a prefixed class name, that may even allow the specificity of the selector to be dropped right down.)

GaryJones (changes requested):

class-tgm-plugin-activation.php
cyberwani: Above request fixed.

class-tgm-plugin-activation.php
cyberwani: It was set to 21 as because if someone wants to add other CSS formatting with low or higher priority. But, I removed the priority.

class-tgm-plugin-activation.php
cyberwani: Changed name ad described above.

class-tgm-plugin-activation.php
cyberwani: Removed double ID, now just using #tgmpa-plugins.

class-tgm-plugin-activation.php
cyberwani: Yes, it may be possible, added provided prefix.

class-tgm-plugin-activation.php
```diff
@@ -2093,6 +2095,19 @@ public static function get_instance() {
 
 			return self::$instance;
 		}
+
+		/**
+		 * Adds CSS to admin head.
+		 *
+		 * @since 2.6.2
+		 */
+		function admin_css() {
```
jrfnl: May I suggest moving this method up a little ? To just above the `get_instance()` method which should be the last in the class ?

class-tgm-plugin-activation.php
```diff
@@ -2741,7 +2756,9 @@ protected function get_row_actions( $item ) {
 		 * @param object $item The current item.
 		 */
 		public function single_row( $item ) {
-			parent::single_row( $item );
+			echo '<tr class="' . esc_attr( 'tgmpa-type-' . strtolower( $item['type'] ) ) . '">';
+			parent::single_row_columns( $item );
```
jrfnl: Please change this to `$this->single_row_columns( $item )` to allow for TGMPA potentially adding that method to the child class.

class-tgm-plugin-activation.php
```diff
@@ -2093,6 +2095,19 @@ public static function get_instance() {
...
 			return self::$instance;
 		}
+
+		/**
+		 * Adds CSS to admin head.
+		 *
+		 * @since 2.6.2
+		 */
+		function admin_css() {
+			echo '<style>
```
jrfnl: I'd suggest starting with a new line, so this code will present aligned in the HTML,

``` php
    echo '
    <style>
```

And maybe indent the css nicely as well ?

class-tgm-plugin-activation.php
```diff
@@ -451,6 +451,8 @@ public function init() {
 			if ( true === $this->has_forced_deactivation ) {
 				add_action( 'switch_theme', array( $this, 'force_deactivation' ) );
 			}
+
+			add_action( 'admin_head',  array( $this, 'admin_css' ) );
```
jrfnl, line 456: Please add a comment above this line to indicate why this hook is there - something along the lines of `// Add CSS for the TGMPA admin page.`

class-tgm-plugin-activation.php
jrfnl: The CSS should only be added when the TGMPA page is being loaded, not for any other admin pages.

``` php
            if ( ! $this->is_tgmpa_page() ) {
                return;
            }
```

jrfnl (changes requested):

GaryJones (approved):

class-tgm-plugin-activation.php
cyberwani: Added.

class-tgm-plugin-activation.php
cyberwani: Yes, it's moved now.

cyberwani: Added.

class-tgm-plugin-activation.php
cyberwani: Changed as above.

class-tgm-plugin-activation.php
cyberwani: Changed.

jrfnl (approved):

PR #632, opened 2016-10-04
Updated cs_CZ translation

mobby2561:
I've updated a few typos + strings consistency.


GaryJones:
@mobby2561 Thanks for this!

Did you change line-endings or something? Your "few typos and string consistency" results in a diff where every single line appears to have been changed.

mobby2561:
I'm sorry but this is my first pull request, so I've just replaced these two files. Should I make a new pull, where I'll replace just the .mo file and exact parts of .po file?

GaryJones:
> Should I make a new pull, where I'll replace just the .mo file and exact parts of .po file?

No need to make a new PR - you can do the `.mo` file and exact `.po` file changes as you suggested, but push (likely rebase and force push) to this branch. If it's easier for you to open a new PR, go with that :-)

PR #637, opened 2016-10-24
Fixed an invalid path preventing the install of the bundled example plugin

mmcev106:


jrfnl:
The example file contains an _example_ path, in this case a path which presumes that TGMPA lives in a `lib` subdirectory and has been unzipped there including the example file (which then would live in `lib/plugins`).

The example file is there to help developers implement TGMPA. Your change would suggest that TGMPA would/should have been unzipped and live in the root of a theme or plugin which is something I for one would definitely strongly advise against.

mmcev106:
Interesting, thanks for the clarification.  The installation directions
just say "Drop the class file somewhere in your theme or plugin hierarchy",
so I just dropped it directly in my theme root.  It took me a minute to
figure out why installing the bundled plugin was not working.  I wonder if
it would be worth changing the installation instructions to include the
suggested "lib" subdirectory.

jrfnl:
I would venture that the place in the theme.plugin directory hierarchy is up to the developer, it really doesn't matter for TGMPA itself. (Though root level is not advised.)
We'd welcome a PR, if you can come up with a good clarification of the installation instructions (readme/example file/gh-pages branch web).

PR #645, opened 2017-02-01
How to activate french translation ?

mod66:
Hi,

I added TGM to my must use plugin and I'm searching to have the sentences in french. I see there is a french traduction but it appears in english on my french wordpress.

Is there a simple way ?

Nothing in the FAQ or config doc.

Thanks.

Simon.


jrfnl:
Not a valid PR, please open an issue if you have a question.

PR #652, opened 2017-03-29
Minor update for WPCS 0.11.0 compliance

jrfnl:


GaryJones (commented):
I'm not a fan of the `// End ...().` statements. Especially adding them in at this stage.

jrfnl:
>I'm not a fan of the // End ...(). statements.

Nor me ;-) We could exclude the rule, but it is one of the core rules. I was a bit ambivalent about doing that as we don't exclude anything else (from core/extra).

Or we could - of course - refactor the code, but that is another matter :sunglasses: .

>Especially adding them in at this stage.

What do you mean ?

GaryJones (approved):

PR #655, opened 2017-04-04
CS: update ruleset with recommended custom properties for improved checking

jrfnl:
Also: use whitelist comment instead of blanket code standards ignore tags.


phpcs.xml
```diff
@@ -1,8 +1,10 @@
 <?xml version="1.0"?>
```
GaryJones, line 1: This file should be called `phpcs.xml.dist`, and `phpcs.xml` should be added to a `.gitignore`.

phpcs.xml
```diff
@@ -14,4 +16,24 @@
...
+		</properties>
+	</rule>
+
+	<rule ref="WordPress.WP.DeprecatedFunctions">
+		<properties>
+			<property name="minimum_supported_version" value="3.7" />
+		</properties>
+	</rule>
+
+	<rule ref="WordPress.WP.DeprecatedFunctions">
```
GaryJones: This looks like an exact copy of the rule above.

GaryJones (changes requested):

phpcs.xml
jrfnl: Could you elaborate why ?

This is the project standard to use and individual contributors should not overrule it with their own `phpcs.xml` file.

phpcs.xml
jrfnl: Sorry, goes to show that I'm working on too many PRs at the same time. The second one should be for `WordPress.WP.DeprecatedParameters`. Adjusting now.

jrfnl: Grr.. and I forgot that that last sniff hasn't been merged yet... oh well.. adding it to a backwing branch with other changes for when WPCS 0.12.0 comes out ;-)

phpcs.xml
GaryJones: Because we're distributing it, but others may want to override this locally with their own PHPCS - maybe to test it against the next version of WPCS or some other reason.

You only need to check https://github.com/squizlabs/PHP_CodeSniffer itself to see `.dist` versions of config files, even if the project has it's own standard to use.

jrfnl: I've made the change as I don't necessarily disagree with your suggestion.

> Because we're distributing it, but others may want to override this locally with their own PHPCS

As it's a library, people shouldn't adjust the code - even for style - when using it.

If it's being used in a larger project, PHPCS will use the `phpcs.xml(.distr)` file in the root - or directories above this - of the larger project anyway.
It will not respect `phpcs.xml` files in directories below the root of the project, so having a `phpcs.xml` in a subdirectory (`/tgmpa`) will not affect the CS run.

> - maybe to test it against the next version of WPCS or some other reason.

The XML file is independent of version - apart from new properties which may be introduced -. What version is used to run it against depends on the version set up locally or in the test environment.

.gitignore
```diff
@@ -0,0 +1,2 @@
+# folders and files to be ignored by git
+phpcs.xml
```
GaryJones: Trailing clear linespace required.

GaryJones (changes requested):

.gitignore
jrfnl: Fixed

GaryJones (approved):

Issue #661, opened 2017-04-19
force_installation !


mohamad-elmi:
Hi There,

As it is described in 'force_activation' field when theme is activated the mentioned plugin gets active but if plugin is not installed so it won’t get activated! I mean I am looking for a field like 'force_installation' in tgm to install and activate the plugin.

Is there any way in this regard?
If no, would you please explain which part of your code handles the installation so I can get my desired state.

Thank you so much for your awesome plugin 👍

jrfnl:
Closing as duplicate of #242, #279, #493 

Installing a plugin should always be the decision of the site admin, not you. Also see:
https://github.com/TGMPA/TGM-Plugin-Activation/issues/279#issuecomment-96334075
https://github.com/TGMPA/TGM-Plugin-Activation/issues/493#issuecomment-155383750

PR #662, opened 2017-04-20
force_installation

mohamad-elmi:
Hi There,

As it is described in 'force_activation' field when theme is activated the mentioned plugin gets active but if plugin is not installed so it won’t get activated! I mean I am looking for a field like 'force_installation' in tgm to install and activate the plugin.

Is there any way in this regard?
If no, would you please explain which part of your code handles the installation so I can get my desired state.

Thank you so much for your awesome plugin :)


jrfnl:
Asked & answered in #661

Issue #669, opened 2017-06-12
Warning: sprintf(): Too few arguments


thimwiz:
I am getting this warning **Warning: sprintf(): Too few arguments** when installing plugins through TGMPA script on **WordPress 4.8**. Is this a known warning with latest version of WordPress?

simrandeep:
I am too facing the same issue. Can you please provide a fix for it?

GaryJones:
Can you be more descriptive about the error message please?

Which plugin or theme is using TGMPA? What version of TGMPA is being used?

thimwiz:
The TGMPA version which is used is **2.6.1**. I am developing this theme.

![tgmpa-4 8-warning](https://user-images.githubusercontent.com/16866500/27022714-d572cd86-4f6c-11e7-90ac-cd17d4b8e2ad.png)

simrandeep:
I have tested it in 4 WordPress themes with TGM v2.6.1 and getting the same error as shown in screenshot above.

al5dy:
I have the same error. The problem appeared after the 4.8v wp.

al5dy:
Bingo! Issue solved. I added some changes in class-tgm-plugin-activation.php

Change:
`$this->upgrader->strings['skin_update_successful'] = %HERE LONG DEFAULT CODE CALLED ERROR%`
to
`$this->upgrader->strings['skin_update_successful'] = __( '%1$s done.' );`

simrandeep:
@al5dy 
Can you attach the full code here?

al5dy:
@simrandeep 
Fix only one function.
https://gist.github.com/al5dy/965d0eefffab32df3b04badbc2274d08

simrandeep:
@al5dy 

Thank you! It fixed the issue :)

thimwiz:
@al5dy Thanks for the fix. Look forward to TGMPA official update.

GaryJones:
Thanks all for the useful feedback - we'll look into this ASAP (as we'll also need to make it work with older versions of WP.

notrealdev:
I have the same error and can not install requires plugin on PHP 5.3
// WP 4.8

jrfnl:
A fix for this issue (#670) has now been merged in and will be included in the upcoming 2.6.2 release.

hellor0bot:
@jrfnl Any updates on the release?

masterbip:
Many thanks @al5dy. Your fix worked!

[...]

PR #670, opened 2017-06-16
Fix "Warning: sprintf(): Too few arguments"

andrezrv:
WordPress 4.8 inserts the "read more" link in the update success message automatically. In this process, `sprintf()` receives only one replacement argument, so PHP throws a warning when two strings are expected to be replaced.

This is the commit in core where the modifications were made: https://github.com/WordPress/WordPress/commit/bab7eb143e3e3651cbb4165a9e904401dd2ed6de#diff-0e5737a4947b2884a63b3a50a59a9d3f.

This issue was already reported in #669, and a fix was proposed. However, I think that checking for the installed version of WordPress is better to keep backwards compatibility.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -3642,8 +3642,6 @@ if ( ! function_exists( 'tgmpa_load_bulk_installer' ) ) {
 							/* translators: 1: plugin name, 2: action number 3: total number of actions. */
 							$this->upgrader->strings['skin_before_update_header'] = __( 'Updating Plugin %1$s (%2$d/%3$d)', 'tgmpa' );
 						} else {
-							global $wp_version;
-
 							/* translators: 1: plugin name, 2: error message. */
 							$this->upgrader->strings['skin_update_failed_error'] = __( 'An error occurred while installing %1$s: <strong>%2$s</strong>.', 'tgmpa' );
 							/* translators: 1: plugin name. */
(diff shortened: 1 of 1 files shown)


GaryJones:
Quick review, but looks sensible. Thanks!

jrfnl:
@andrezrv The travis error had nothing to do with your PR and has been fixed in the mean time. Would you mind rebasing the PR so we can get a passing build ? Thanks!
I'll test the PR after that and will merge it if I don't find any problems with it (I don't expect any issues).

andrezrv:
@jrfnl Done, and I added one more commit to fix the end of line issue reported by Travis.

jrfnl (approved):
I've reproduced the issue on WP 4.8 and tested the fix on WP 4.6, 4.7 and 4.8 and found it all working as expected.

I made one small tweak to the code, replacing the `global` statement with using the locally cached version of the `$wp_version` variable for efficiency.
The variable is cached in the TGMPA constructor: https://github.com/TGMPA/TGM-Plugin-Activation/blob/develop/class-tgm-plugin-activation.php#L262

GaryJones (approved):

Sharifur:
change this function to this 

https://gist.github.com/Sharifur/cdf71a63fb402eb8f38f6fbe03e0522b

Ninetheme:
> change this function to this
> 
> https://gist.github.com/Sharifur/cdf71a63fb402eb8f38f6fbe03e0522b

thank you!

PR #674, opened 2017-06-29
editorconfig: Be a little more specific

GaryJones:
Use file extensions to be a little more specific with configuration.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,13 +9,19 @@ root = true
 [*]
 charset = utf-8
 end_of_line = lf
-indent_size = 4
-tab_width = 4
-indent_style = tab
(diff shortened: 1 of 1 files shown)


jrfnl (commented):
LGTM, but as I don't use `.editorconfig` myself, I can't properly gauge the effect.

GaryJones:
This is the latest version of the standard `.editorconfig` I put in all my WP plugins and themes - it has no effect on the code logic itself, and only adjusts the code style in editors that support EditorConfig. It's a safe PR.

jrfnl:
Figured as much, I'm happy for it to be merged.

jrfnl (approved):

Issue #687, opened 2017-08-23
[Question] Wp Favs Integration


timersys:
Hi Guys, just wanted to ask before getting my hands dirty if you would accept a PR to add wp favs integration to the TGM plugin so users could add a [simple token key](https://wpfavs.com/wpfavs/editors-choice) instead of the bundled plugins zips.
That way theme authors could considerably reduce the size of themes.

I'm currently using a custom coded version of TGM in  the [Wp Favs plugin](https://wordpress.org/plugins/wpfavs/) but adding native support to the plugin will benefit lot of users.

That token mentioned early return direct download urls that can be passed to TGM. Example from the list above can be found in https://wpfavs.com/api/v1/wpfav/3b871fb19229696af32307da785d66e0

Regards

GaryJones:
If you can do the PR, and the end result is that it doesn't negatively affect anyone not using wpfavs.com, then we can definitely consider merging it. Thanks!

timersys:
Great, that's all I needed to hear. Thanks

PR #690, opened 2017-08-26
Wp Favs integration

timersys:
Hi Guys as commented on #687 here is the pull request with a simple integration of Wp Favs. 
I just added one new function to grab the plugins from the API and updated the example.php with a working token example.

Thanks!

ps: Why you use themes as default screen for installing plugins? Just curious

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -2200,40 +2200,44 @@ if ( ! function_exists( 'tgmpa' ) ) {
 	}
 }
 
-if( ! function_exists( 'tgmpa_wpfavs_plugins' ) ) {
+if ( ! function_exists( 'tgmpa_wpfavs_plugins' ) ) {
 	/**
 	 * Custom function that grabs plugins from https://wpfavs.com
 	 * from given token
+	 *
 	 * @since 2.6.2
 	 *
-	 * @param string $token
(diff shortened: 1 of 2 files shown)


jrfnl:
Hi @timersys, I haven't looked at the code yet, but I like the idea of this.
However, users can also have a favorites list on wordpress.org. So I'm wondering if this PR would be accepted, if the favorites list on WP.org shouldn't also be supported.
Supporting an external service, but not a WP native implementation feels arbitrary.

Also, as this is a new feature, I think it should go into a minor release - in contrast to a patch release, like the 2.6.2 version we're currently preparing for -, so I'm tagging this with 2.7.0. /cc @GaryJones 

As a side-note: you may want to have a look at the Travis output. The codestyle errors will need to be fixed before this PR can be considered: https://travis-ci.org/TGMPA/TGM-Plugin-Activation/jobs/268774079#L685-L760

timersys:
Hey @jrfnl , I could add the native wp.org favorites support. I already got that bit of code so no problem with that.
I will review everything and update accordingly.

timersys:
So I added the wp favorites feature. Take in consideration that I added these 2 features as separate functions and didn't integrate them in the main class as I wasn't sure you wanted to go that route.

PR #707, opened 2017-11-03
WP-Translations updates

fxbenard:
Our [Team](https://wp-translations.pro/team/) and all our contributors are happy to support [your project](https://www.transifex.com/wp-translations/tgm-plugin-activation/dashboard/)


jrfnl:
@fxbenard This is awesome ! Thank you so much. It may be a little while before we get round to reviewing & merging this, but please know that we really appreciate your support and these translations.

jrfnl:
Hi @fxbenard Again, sorry for the delay before I had a chance to look at this properly and thanks again for your support.

Now, there are a couple of things I'd like to discuss with you:
1. Before anything else, the PR needs to be rebased. Sorry about that.
2. You appear to have updated the `.pot` file and the file headers (and line nrs) of all existing translation files.
    Some of the information now included in the file headers is incorrect.
    Also, for a number of translations, it looks like the actual translations haven't been updated at all, just the `.po` file synced with the updated `.pot` file. This is undesirable as it makes reviewing the PR a lot more difficult.

    Generally speaking, so far, we would do an update to the `.pot` file just before a release and update the `.po` files to match. We would then invite the translators who have been previously involved to review the updated files for their language and add/update translations where necessary before the new version of TGMPA is released.

    If you think that process can be improved, please let me know and I'd be happy to discuss it.
3. The `ca` (Catalan) and `nb_NO` (Norwegian Bokmal) translations included in the PR are very incomplete and not suitable yet for inclusion in the project. As a general rule of thumb, at least 90% of the text phrases should be translated before a translation will be included.
    The `zh_CN` (Mandarin Chinese) translation and the `da_DK` (Danish) translation are both much closer to that mark, but also would not make it yet.

PR #715, opened 2018-01-13
Ensure composer.lock stays out of version control

GaryJones:
This is a library package, not an application.

See https://blog.martinhujer.cz/17-tips-for-using-composer-efficiently/#tip-%236%3A-put-%60composer.lock%60-into-%60.gitignore%60-in-libraries

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
-# folders and files to be ignored by git
-phpcs.xml
+/vendor/
+/composer.lock
+/phpcs.xml
+/.phpcs.xml


jrfnl (approved):
It should be noted that, at this moment, TGMPA does not require any dependencies, so there would never be a `vendor` directory.

All the same, I agree this is a good change as it will protect against issues in the future if dependencies would be added.

jrfnl:
@GaryJones Tip 13 of the article looks like a great suggestion as well. Do you intend to pull that too ?

GaryJones:
Wasn't intending to, but we can. We'd just need to get the SPDX identifier updated, depending on the version of Composer that our Travis setup is running.

jrfnl:
@GaryJones I've just run some tests:

* On Composer 1.5.0, using the newer license annotation will cause validation to fail.
* On Composer 1.6.0+, using the old license annotation will cause validation to fail.

In both cases, it's only a warning, so validating with `composer validate --no-check-all` - note: without the `--strict` - exits without errors on both versions I tested.

PR #716, opened 2018-01-13
Use SPDX License List 3.0 identifier

GaryJones:
See https://spdx.org/news/news/2018/01/license-list-30-released

Composer 1.6 considers `GPL-2.0+` to be invalid.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
   "description": "TGM Plugin Activation is a PHP library that allows you to easily require or recommend plugins for your WordPress themes (and plugins).",
   "homepage": "http://tgmpluginactivation.com",
   "keywords": ["wordpress","plugins", "theme", "library", "activation"],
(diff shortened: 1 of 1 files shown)


jrfnl (approved):

jrfnl:
Just one question: Composer 1.6 has only been out less than two weeks. Do older Composer versions recognize the new license annotation correctly ?

[Edit] Based on the [changelog](https://github.com/composer/composer/releases/tag/1.6.0), I'd guess not.
For now, it looks like the "old" annotation is deprecated, not dropped completely yet, so maybe we should wait a little longer before making this change.

GaryJones:
Anyone using an older version, shouldn't experience any problems unless they try to run `composer validate` on TGMPA. There's no harm in leaving it for a while, but equally, I can't see any benefit either.

PR #720, opened 2018-01-14
Add a `.gitattributes` file

jrfnl:
This PR adds a `.gitattributes` file to keep the archives GH creates of the repo clean of development related files.
People using Composer can still get these files in their setup if they really want to, by using `--prefer-source`.

Refs:
* [Reddit: I don't need your tests in my production](https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production)
* [Blog: I don't need your tests in my production](https://blog.madewithlove.be/post/gitattributes/)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,28 @@
+#
+# Exclude these files from release archives.
+# This will also make them unavailable when using Composer with `--prefer-dist`.
+# If you develop for TGMPA using Composer, use `--prefer-source`.
+# https://www.reddit.com/r/PHP/comments/2jzp6k/i_dont_need_your_tests_in_my_production
+# https://blog.madewithlove.be/post/gitattributes/
+#
+/.editorconfig export-ignore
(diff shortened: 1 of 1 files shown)


GaryJones:
I'd also consider adding the following, none of which are needed on production deployments:

- `/plugins/tgm-example-plugins.zip`
- `/CHANGELOG.md`
- `/CONTRIBUTING.md`
- `/example.php`
- `/README.md`
- `/LICENSE.md`

jrfnl:
For production deployment, they are not needed, but they ought to be in a release archive as the library should be distributed as a complete package.

The only file from your list which I clearly missed when preparing the PR is the `Contributing.md` file. Added that in a new commit.

The `changelog.md` file is debatable and could possibly be left out, but everything else should be included IMO.

As a side-note: not including the license file in the package, AFAIK invalidates the license (at least for most license types)...

GaryJones (approved):

PR #749, opened 2018-07-08
Add has_notices on plugin level

### LINKED ISSUE #748: How to display notice for required plugins only?
We have a theme that uses different plugins depending on install. 
2 plugins are always required other 5

```
BudddyPress
bbPress
WooCommerce
LayerSlider
Slider Revolution
```

are only recommended 

How can we display notice only for required plugins?

danyj:
This PR sets the `has_notice` on plugin level so that we can disable the notices for recommended plugins if we need to. Fixes #748 

Usage 

		'bbpress'=> array(
			'name'      => 'bbPress',
			'slug'      => 'bbpress',
			'required'  => false,
			'force_activation'   => false,
			'has_notices' => false // This plugin will not be visible in the TGMPA notice
		),

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1153,7 +1153,6 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				if ( $this->is_plugin_active( $slug ) && false === $this->does_plugin_have_update( $slug ) ) {
 					continue;
 				}
-				
 				if ( false === $plugin['has_notices'] ) {
 					continue;
 				}
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1153,6 +1153,10 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				if ( $this->is_plugin_active( $slug ) && false === $this->does_plugin_have_update( $slug ) ) {
 					continue;
 				}
+				
+				if(isset($plugin['has_notices']) && false === $plugin['has_notices']){
+					continue;
+				}
 
 				if ( ! $this->is_plugin_installed( $slug ) ) {
 					if ( current_user_can( 'install_plugins' ) ) {
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1154,7 +1154,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 					continue;
 				}
 				
-				if(isset($plugin['has_notices']) && false === $plugin['has_notices']){
+				if(false === $plugin['has_notices']){
 					continue;
 				}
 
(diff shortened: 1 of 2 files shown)


jrfnl (changes requested):
Hi @danyj Welcome and thanks for your first PR!

I realy like the idea of this change.

As we try to make sure the code in TGMPA is high-quality, for this PR to be acceptable a couple of changes are needed:
1. The code needs to comply with the WordPress Coding Standards - see the Travis output for the issues.
2. TGMPA uses the principle that all plugin options are always set. To comply with this:
    - Add the new option to the `$defaults` in the `register()` method: https://github.com/TGMPA/TGM-Plugin-Activation/blob/2732e40a65738604669d2c63079fd4dcfb430c97/class-tgm-plugin-activation.php#L1399-L1409
    - Add validation for the input in the same method - see line1421 for an example: https://github.com/TGMPA/TGM-Plugin-Activation/blob/2732e40a65738604669d2c63079fd4dcfb430c97/class-tgm-plugin-activation.php#L1420
    - You can now remove the `isset($plugin['has_notices'])` from the condition you added.
3. It may be useful to add an example using this new option to the `example.php` file.
4. A sister-PR is needed for the `gh-pages` branch to adjust the `configuration` documentation: http://tgmpluginactivation.com/configuration/

Please let me know if you need help with any of this and I look forward to seeing this get into TGMPA.

jrfnl:
@GaryJones As this adds a new feature - do you agree that if this is merged in, the next version will need to be called `2.7.0` (instead of `2.6.2`) ?

danyj:
@jrfnl on it ,  going trough  suggested changes.

danyj:
not sure why travis is failing

danyj:
> A sister-PR is needed f

where is that git ?

danyj:
@jrfnl  please check the travis , all those lines were there and are untouched. I simply pulled the git and changed 4 lines and they do not seem to be failing.

danyj:
Checked travis , all the lines mentioned in both files were there before my PR.

danyj:
finally :)

jrfnl:
@danyj Thanks for all the changes. I'll try to have another look later today.

> where is that git ?

Regarding the sister-PR, it is in this same repo, just on another (orphan) branch: https://github.com/TGMPA/TGM-Plugin-Activation/tree/gh-pages

danyj:
All set. Submitted PR for it.

jrfnl (commented):
@danyj All looks good to me.

There's just one thing I'm wondering: should `has_notices` always be allowed/respected or only when `required` is `false` ?

If a plugin is required and not active, not showing the notice could easily cause the fact that the plugin is deactivated to go unnoticed, which - depending on the quality of the theme/plugin which requires it - could break sites with a WSOD.

What do you think ?

@GaryJones Have you got an opinion on this ?

If we would only allow it for `recommended` plugins, the `if()` within the `notices()` method would need to check the `required` setting as well + the explanation text in `example.php` would need to mention this too.

danyj:
> or only when required is false

good point , I think only when required is `false `because `required `  `true` should always show notice. 

@GaryJones  @jrfnl , let me know and I can make changes if you agree

Issue #752, opened 2018-07-13
Mo files are not well generated


Tofandel:
Hi,

I noticed today that the mo files are not well compiled (at least for the french language)

The singular and plurals are concatenated together so I checked the po file and everything is in order, I recompiled it with poedit and then the translations started displaying correctly

I can provide the newly compiled mo file for Fr and other languages that might have the same issue as well if needed.

To reproduce => install a plugin with the latest version of tgmpa, requiring another plugin, with wordpress in french,
And notice that the string for "Begin to install the plugin" is : "Commencer à installer l'extensionCommencer à installer les extensions" instead of just "Commencer à installer l'extension"

franzarmas:
Indeed, I have the same problem for the Spanish language (es_ES).

I also did the check with Poedit and after the compilation the translations start to work.

Please solve this in the future version.

jrfnl:
@Tofandel @franzarmas We normally update the .po(t) files before each new release and regenerate the translation .mo files after translators have submitted any updates they want to provide. I'd be very happy if you'd be willing to test those files before we release the next version to prevent this issue.

Tofandel:
@jrfnl What tool do you use to regenerate them ?

jrfnl:
@Tofandel If I remember correctly I used PoEdit originally, but the last batch may have been generated using https://github.com/phpmo/php.mo
It's been a while, so not entirely sure at the moment.

cdmmultimedia:
We are having the same problem (i.e. weird plurals and singular sentences cobbled up together, and some translations falling back to the default). Regenerating the .mo files from the .po files fixed the issue.

cdmmultimedia:
The library that you're mentioning seems to be no longer maintained. Do you need help with getting a reliable CI-enabled generation of the .mo files ?

jrfnl:
@cdmmultimedia Help setting that up or even just documenting how we can do this easily in bulk would be awesome!

cdmmultimedia:
@jrfnl We'd be happy to look into it :)

jrfnl:
@cdmmultimedia Much appreciated! When you do, please keep in mind that maintainers may be on varying OS-es, i.e. both Linux/Mac as well as Windows will need to be supported.

PR #754, opened 2018-08-01
Add a class_exists to check if plugin is active (like is callable)

Tofandel:
This PR adds an option 'class_exists' to check if plugin is active (like the is_callable option)

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -1406,6 +1406,7 @@ if ( ! class_exists( 'TGM_Plugin_Activation' ) ) {
 				'force_deactivation' => false,   // Boolean.
 				'external_url'       => '',      // String.
 				'is_callable'        => '',      // String|Array.
+				'class_exists'        => '',      // String.
 			);
 
 			// Prepare the received data.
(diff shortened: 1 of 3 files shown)


jrfnl:
@Tofandel Hi Adrian, thank you for your willingness to contribute. Could you please explain what this adds what you couldn't already do with `is_callable()` ?

Every class should have at least one public method otherwise it cannot be instantiated. Using that method with is_callable, as in passing `someClass::SomeMethod` as a value to the array, should work already.

Tofandel:
@jrfnl Hello, well is callable will work only with public static functions, already instantiated objects's private methods (which you cannot do if the plugin is inactive) or public methods [**BUT NOT THE CONSTRUCTOR**](http://php.net/manual/en/function.is-callable.php#example-6501)

So this would add the possibility to check if a class exists which you cannot currently effectively do with is_callable if only a constructor is available and no other public methods within the class (which is usually the case with well structured plugins)

Thanks

Tofandel:
We could also argue that for an hypotetical plugin which only add_filters and actions with anonymous functions and have no class defined or function whatsoever it would be very hard to check that they do exist, but they may define a constant with their version or whatever.
Then an is_defined option would also make some sense (even if this is very unlikely that there is not a single function or class) but that would give the entire flexibility of checking whether or not a plugin indeed exists with different methods that can fallback on each other since the condition is made with an 'or' statements

Since it doesn't break anything and just gives some flexibility I wouldn't see why not include these options
What do you think ?

By the way I didn't had any answer on the ['mo files are not well generated'](https://github.com/TGMPA/TGM-Plugin-Activation/issues/752) issue, should I make a PR with all the mo files regenerated ?

Edit: Added is_defined to the PR following the logic

Issue #794, opened 2020-08-24
Allow add_submenu_page for WordPress.org themes via Generator


acosmin:
@jrfnl We've started allowing `add_menu_page` & `add_submenu_page`[ in themes on wp.org](https://make.wordpress.org/themes/2020/08/02/new-admin-menu-page-requirement/) and we've noticed that `parent_slug` uses only `add_theme_page` via TGMPA Generator. This piece of code https://github.com/TGMPA/TGM-Plugin-Activation/blob/2d34264f4fdcfcc60261d490ff2e689f0c33730c/class-tgm-plugin-activation.php#L740-L742 should be allowed now. I guess the Generator needs to be changed.

Thank you!

jrfnl:
@acosmin Ripping that out via the generator was done at the explicit request of the wp.org Theme team, so yes, if you now want to start allowing it, the generator will need to be updated. PRs welcome.

acosmin:
Sure :) could you please point me to the Generator repo?

acosmin:
I think I found it :) https://github.com/TGMPA/TGM-Plugin-Activation/blob/gh-pages/js/tgmpa.js I'll take it for a spin

PR #795, opened 2020-08-26
Removed .org restriction

acosmin:
@jrfnl  Regarding #794

I noticed that themeforest is also using the `add_theme_page` so I left it in a conditional.

The minified version is generated by javascript-minifier.com

Yui compressor is broken on their end and I couldn't minify the file.

I also checked the min version and no JS issues were found. All the zip files are generated as intended.


jrfnl:
@acosmin Thanks for this PR. I want to start up a local instance of GH Pages to test it, especially as the GH Pages version has changed a lot since we last deployed, but the code changes themselves look good.

I'll try to get round to the testing soon.

PR #805, opened 2021-04-27
Allow Per-Instance Configuration

### LINKED ISSUE #804: Improve Separation of Concerns
## The Problem
The functionality of this library is very useful. It allows developers to quickly specify the plugin dependencies of their plugin or theme, and provide users with an intuitive way to quickly install these dependencies.

Unfortunately, having TGMPA in a plugin introduces a risk of conflict with other plugins that use TGMPA. This is especially true with string customization, as one plugin's strings would overwrite TGMPA string configuration done by another plugin.

## Suggested Solution
The approach I took to ensure that my TGMPA instance does not conflict with others (apart from different versions of this same lib, which is [kinda inevitable for the most part][package-management-wordpress]) is to have a separate instance of `TGM_Plugin_Activation` for my plugin. This instance does not overwrite the global instance, and thus all my configuration is preserved separately. This results in an additional notice and menu item per plugin [that uses a separate instance], which is arguably better: each plugin or theme has a different list of dependencies, and updating all dependencies of all plugins in one go is not necessarily what the user wants.

However, I've encountered the following issues:

1. On the "Install Plugins" page, the lists for both plugins to update and install are empty, even if the notice is displayed correctly. This is because `TGMPA_List_Table`, much like all the other parts of this library, [uses the _global_ instance of `TGM_Plugin_Activation`][1], not the one I have configured.

	This can easily be solved by dependency-injecting `TGM_Plugin_Activation` into the list.

2. Custom strings have no effect, at least on the notice. This is because the `TGM_Plugin_Activation` instance is configured with default strings _after_ its initialization, and therefore normal configuration, is complete. So, if I configure my instance with strings right after instantiating it, those strings would be overwritten later with defaults - because the default strings are [assigned][2] on [init][3].

	This can easily be solved by assigning default strings in the constructor, and then letting initialization overwrite them (as defaults should work IMHO), subsequently using that configuration as is. Arguably, all configuration should be done during initialization, preferably by dependency-injecting it, or less favourably by assigning defaults as close as possible to the constructor.

[1]: https://github.com/TGMPA/TGM-Plugin-Activation/blob/develop/class-tgm-plugin-activation.php#L2277
[2]: https://github.com/TGMPA/TGM-Plugin-Activation/blob/develop/class-tgm-plugin-activation.php#L339
[3]: https://github.com/TGMPA/TGM-Plugin-Activation/blob/develop/class-tgm-plugin-activation.php#L277
[package-management-wordpress]: https://inpsyde.com/en/package-management-in-wordpress-introduction-solutions/

XedinUnknown:
Fixes TGMPA/TGM-Plugin-Activation#804.

In general, this change simply shifts things to where they should be in order to properly configure multiple instances of TGMPA.

- TGMPA string configuration now overridable during configuration (initialization).
- Update/install/activate lists reflect requirements of the correct configuration.
- Updating/installing/activating in bulk works on the correct configuration.

### DIFF
--- a/class-tgm-plugin-activation.php
+++ b/class-tgm-plugin-activation.php
@@ -3196,602 +3196,564 @@ if ( ! class_exists( 'TGM_Bulk_Installer_Skin' ) ) {
 	}
 }
 
-/**
- * The WP_Upgrader file isn't always available. If it isn't available,
- * we load it here.
- *
- * We check to make sure no action or activation keys are set so that WordPress
- * does not try to re-include the class when processing upgrades or installs outside
- * of the class.
- *
- * @since 2.2.0
- */
-add_action( 'admin_init', 'tgmpa_load_bulk_installer' );
-if ( ! function_exists( 'tgmpa_load_bulk_installer' ) ) {
+if ( ! class_exists( 'TGMPA_Bulk_Installer' ) ) {
+
+	require_once ABSPATH . 'wp-admin/includes/class-wp-upgrader.php';
+
 	/**
-	 * Load bulk installer
+	 * Installer class to handle bulk plugin installations.
+	 *
+	 * Extends WP_Upgrader and customizes to suit the installation of multiple
+	 * plugins.
+	 *
+	 * @since 2.2.0
+	 *
+	 * {@internal Since 2.5.0 the class is an extension of Plugin_Upgrader rather than WP_Upgrader.}}
+	 * {@internal Since 2.5.2 the class has been renamed from TGM_Bulk_Installer to TGMPA_Bulk_Installer.
+	 *            This was done to prevent backward compatibility issues with v2.3.6.}}
+	 *
+	 * @package TGM-Plugin-Activation
+	 * @author  Thomas Griffin
+	 * @author  Gary Jones
 	 */
-	function tgmpa_load_bulk_installer() {
-		// Silently fail if 2.5+ is loaded *after* an older version.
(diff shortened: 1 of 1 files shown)


XedinUnknown:
The build seems to be failing due to misconfigured CI: some packages require Composer v1, while Composer v2 is used on CI.

jrfnl:
As per https://github.com/TGMPA/TGM-Plugin-Activation/issues/804#issuecomment-827580580:

> TGMPA is _single instance_ by design. Changing that is not open for discussion.

I suggest closing this PR.

XedinUnknown:
As I have explained, nothing should change for the current users with this PR. Everyone should be able to continue using the single instance. Also, a class is not single-instance by design; the single-instance-ness is incidental, in that you can force a class to only have a single instance through high coupling etc, but this has no benefits while simply limiting consumers for no apparent reason. Your _component_ (a part of an _application_, like a plugin or this lib) surely can use only a single instance of something, and can provide to its consumers ways to maintain consistency through the single instance - like by initializing an object, caching it, and always retrieving from cache. In this case, the `tgmpa` global is being used for that cache. So your component remains the same in the sense that it continues using only a single instance, and all of the consumers that follow your docs will thus use a single instance.

Given the above, I do not understand the relevance of your comment, sorry. Would you mind explaining exactly what the problem is, e.g. where in any code in this lib, including my PR, you see that it is no longer using a single instance?

